PR #14, opened 2021-06-16
Migrate GitHub pr

Irma12:


.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: yarn install
+
+      - name: Lint
+        run: yarn lint
+
+      - name: Run tsc
+        run: yarn tsc
+        
+      - name: Check updates in dependencies
+        run: yarn backstage-cli versions:bump
```
martina-if, line 43: I think I would remove this from this step or we would be building something that's not what we have in the PR.

Then I'd leave this step only in the nightly workflow.

@Xantier  opinions?

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          start: yarn dev
+          wait-on: 'http://localhost:3000'
+          project: ./packages/app
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  package-update:
+    needs: build
```
Xantier, line 54: Should the cypress step be run after this package updates?

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,69 @@
...
+        run: yarn tsc
+        
+      - name: Check updates in dependencies
+        run: yarn backstage-cli versions:bump
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v2
+        with:
+          start: yarn dev
+          wait-on: 'http://localhost:3000'
```
martina-if, line 49: Nice, I didn't know about this `wait-on` :+1:

Xantier (approved):

.github/workflows/on-pull-request.yml
```diff
@@ -0,0 +1,69 @@
...
+      uses: technote-space/create-pr-action@v2
+      with:
+         EXECUTE_COMMANDS: |
+           npx npm-check-updates -u --packageFile package.json
+           yarn install
+           yarn backstage-cli versions:bump
+           yarn upgrade
+         COMMIT_MESSAGE: 'chore: update npm dependencies'
+         COMMIT_NAME: 'GitHub Actions'
+         COMMIT_EMAIL: 'roadie@roadie.io'
```
martina-if, line 67: should it be `bot@roadie.io` instead?

martina-if (commented):
Mostly Looks Good To Me, I just have a question around bumping versions when we open a PR. I think it's better to do that as a separate step.

.github/workflows/on-pull-request.yml
Irma12: We do have update on line 42, before cypress run, I see now this is a bit confusing. This job is actually for PR creation, maybe renaming it would be good for later @Xantier

.github/workflows/on-pull-request.yml
Irma12: I was thinking bumping this in should be good in order to make sure cypress tests are passing for packages if we update versions

.github/workflows/on-pull-request.yml
Irma12: @iain-b opened my eyes for this one :)

.github/workflows/on-pull-request.yml
Xantier: I think I might agree with Martina. We should be able to expect that PRs are up to date with `main` and `main` should have a workflow to bump versions. Therefore this wouldn't be needed on PRs since they get correct versions numbers from `main` already

Irma12: I see, we do have this in bump-version.yml already, so it is reasonable to remove it from here then :)

PR #28, opened 2021-06-18
Workflow setup

Irma12:


.github/workflows/e2e-tests.yml
```diff
@@ -49,17 +48,16 @@ jobs:
...
     - name: Update npm packages
       uses: technote-space/create-pr-action@v2
       with:
          EXECUTE_COMMANDS: |
-           npx npm-check-updates -u --packageFile package.json
            yarn install
            yarn backstage-cli versions:bump
-           yarn upgrade
-         COMMIT_MESSAGE: 'chore: update npm dependencies'
+         COMMIT_MESSAGE: 'Update dependencies'
```
Xantier, line 60: Maybe add something like "Automated commit by something something"

Xantier (approved):

.github/workflows/main.yml
```diff
@@ -0,0 +1,45 @@
...
+  publish-packages:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+    - name: Publish packages
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: '0'
+    - run: |
+        lerna version patch
```
iain-b, line 41: This is ok for now but just thinking will we always need to bump every package ? (perhaps someone bumps a package manually). Or can we update and publish a single package ? (maybe a contributor is having an issue and just wants to fix a specific plugin).
Also, what if a publish fails (probably shouldn't happen but say we've already published the version)?
Finally, how can we do minor/major releases? 
No need to tackle these questions with this PR I'm just thinking out loud.

iain-b (approved):

.github/workflows/main.yml
Irma12: I think it is fair point @iain-b. Shouldn't lerna detect only those packages then? Ones that are changed. For minor and major bump versions, maybe for now we could manually do it for the packages when they are modified. It will however increment patch but I don't think it is such big of an issue. Might be wrong.

Regarding what happens if it fails, hmm if we use tags, maybe if the workflow fails we can follow steps as described here https://github.com/lerna/lerna/issues/455#issuecomment-792821661

PR #35, opened 2021-06-23
Modify workflow

Irma12:


.github/workflows/weekly-check-versions.yml
```diff
@@ -4,15 +4,39 @@ on:
...
+      - name: Create Pull Request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN  }}
+          commit-message: Update dependencies
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          signoff: false
+          branch: update-dependecies
```
iain-b, line 28: Can we have concurrent PRs open with this ? perhaps we should put the date and perhaps time in the branch name?

iain-b (approved):
Looks good to me, just a thought on branch names but it's a future improvement

.github/workflows/weekly-check-versions.yml
Irma12: I am not sure about concurrent PRs, what would be use case for those? Yeah, that's good idea, I will put the date in branch name :) @iain-b

iain-b: I'm thinking we'll end up wanting to run this more regularly than once a week and perhaps we haven't merged a PR then the creation of the next PR fails as the branch already exists.

Irma12: Good point, I think it will just update the existing branch and treat that as one PR.

PR #66, opened 2021-07-05
Post-merge fix

Irma12:
Using lerna can be tricky, especially when it comes to versioning so we may use conventional commits in addition to this workflow for publish.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -5,7 +5,7 @@ on:
       - main
 jobs:
   publish:
-    name: npm-publish
+    name: publish
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Xantier (approved):

.github/workflows/publish.yml
```diff
@@ -4,23 +4,36 @@ on:
...
+  
+    - uses: actions/setup-node@v1
       with:
         node-version: 14.x
-    - run: yarn install
-    - run: yarn tsc
-    - run: yarn build
-    - run: |
-        git config --global user.email "bot@roadie.io"
+        registry-url: 'https://npm.pkg.github.com/'
```
iain-b, line 18: Why publish to github out of curiosity ?

Irma12: Oh, this shouldn't be here, good catch. I was looking at some solutions so this is leftover. However, I will decline this PR as we won't be doing this now @iain-b

PR #93, opened 2021-08-03
Bumped a few dependencies and updating to use later `material-ui`

benjdlambert:
I think some peoples `yarn backstage-cli versions:bump` is failing due to some roadie dependencies being out of date, so I've tried to update things to latest version :)


Xantier:
Thanks @benjdlambert, this is a duplicate oh #89 which is running through the CI pipeline now and will be merged in a bit.

benjdlambert:
Ah ok - sorry - should have checked first :)

teppichKay:
@benjdlambert Hey I think your PR isn't a fully duplicate one with #89 (see my comments there)

benjdlambert:
@teppichKay I think that the end result should still be the same, as it won't bring in two versions if they match the same semver, you can see that the `yarn.lock` did bring in the correct version without the change. I just did it as a backup, but it should probably for consistency be changed too

teppichKay:
yeah you're right. Thanks!

PR #136, opened 2021-08-11
Attempt to fix System limit for number of file watchers reached

punkle:
The e2e are failing often with the error System limit for number of file watchers reached.

This is an attempt to fix it.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -40,7 +40,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
-          start: CHOKIDAR_USEPOLLING=true yarn dev
+          start: CI=true yarn dev
           wait-on: 'http://localhost:3000'
           project: ./packages/app
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-tests.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
-          start: yarn dev
+          start: CI=true yarn dev
```
dtuite, line 43: Would it be better to set `CI=true` for the whole workflow rather than just this one command?

dtuite (approved):

.github/workflows/e2e-tests.yml
punkle: Maybe yeah. I'm stabbing in the dark. :/

punkle: Its already set. :(

dtuite: So it is.

PR #142, opened 2021-08-11
Update Jira User-Agent Dummy Value To Be More Intuitive

aravindparappil46:
The previous value `MY-UA-STRING` might cause confusion on whether to replace that string with an actual browser user agent string. Any non-whitespaced string should work here.

Also made some minor changes to the same `README.md`

[Related Discord conversation](https://discord.com/channels/760879902391009330/760879902391009333/874014966242504706)


martina-if:
https://app.clubhouse.io/larder/story/3530

dtuite (approved):

Issue #199, opened 2021-08-27
ArgoCD plugin expects missing JSON field


sam-robson:
I've been having some problems with the Argo plugin when applications are syncing. 
Specifically I am receiving a `remote data validation failed` error as the JSON being validated lacks the [finishedState field on operationState](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/backstage-plugin-argo-cd/src/types.ts#L13).

<img width="1064" alt="roadie-argo-plugin-error" src="https://user-images.githubusercontent.com/20035092/131100602-781c2aa1-f549-45b1-9a84-5405507e2ba3.png">

My first thought is that it makes sense that the value is not present, as the syncing operation hasn't finished. If so, it shouldn't be expected. I will write up a quick PR to remove this.

martina-if:
Ch3877

PR #200, opened 2021-08-27
fix: replace finishedAt with startedAt in Argo

sam-robson:
Operations have always started, but not always finished.
https://github.com/RoadieHQ/roadie-backstage-plugins/issues/199

### DIFF
--- a/plugins/backstage-plugin-argo-cd/src/types.ts
+++ b/plugins/backstage-plugin-argo-cd/src/types.ts
@@ -10,7 +10,8 @@ export const argoCDAppDetails = t.type({
       status: t.string,
     }),
     operationState: t.type({
-      finishedAt: t.string,
+      startedAt: t.string,
+      finishedAt: t.union([t.string, t.undefined]),
     }),
     history: t.array(t.type({
       id: t.number,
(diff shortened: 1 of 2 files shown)


Xantier:
Thanks for the contribution @sam-robson! I'll spin this up on our ArgoCD instances and take a look && merge eventually

plugins/backstage-plugin-argo-cd/src/types.ts
```diff
@@ -10,7 +10,7 @@ export const argoCDAppDetails = t.type({
       status: t.string,
     }),
     operationState: t.type({
-      finishedAt: t.string,
+      startedAt: t.string,
```
Irma12, line 13: Even though I have nothing against adding startedAt property, I would keep finishedAt in a form of:
finishedAt: t.union([t.string, t.undefined])
(The reason behind this is we use finishedAt in order to display 'Last synced' data) https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/backstage-plugin-argo-cd/src/components/ArgoCDDetailsCard.tsx

iain-b (approved):
@Irma12 Your comment was addressed so this seems good to go ?

Irma12:
@sam-robson Thank you for the contribution once again 🎉

Irma12 (approved):

PR #207, opened 2021-09-03
Change severity colors

Irma12:
Preview on dark and light background for severity level

Screenshot 2021-09-03 at 22 04 13
Screenshot 2021-09-03 at 22 04 25


dtuite (approved):
Wow that was fast!

PR #208, opened 2021-09-07
per cluster argo support

### LINKED ISSUE #193: ArgoCD instance per cluster
We have one argocd instance per cluster. The proxy solution mentioned doesn't seem scalable to us. We'd also need to add a path on catalog files for this to work properly.

The current kubernetes plugin finds an app across all clusters based on selectors. Could this plugin do something similar instead of having a proxy path added to each and every catalog-info file?

I started writing code to take over the argocd proxy URL on the backend to redirect the traffic based on what cluster an app resides on, but argocd plugin doesn't send cluster info making this harder and we'd rather work with you to make a solution than working in a silo.

All our argocd instances have URLs based on a pattern. Given the cluster name, we can determine the argocd instance URL.

karlhaworth:
This is to continue the discussion that I had with @Xantier . This is only the frontend. Once we agree on a approach, if it's something willing to be accepted or continued to be worked on, then we can continue to the backend that we built, but have not yet published as open source.

We install Argo CD on every cluster. We chose the approach of using the existing kubernetes block in the app-config.yaml files so that we did not duplicate information. Since each argo cd instance uses different tokens, we're using a username and password across all clusters and generating the token using the backend we created (to be discussed). That user can be synced with LDAP or something similar using argocd native functionality, but we did not want to keep track of many different tokens per cluster. This makes it dynamic and when we rotate credentials we can update in a single spot.

This solution unfortunately tightly couples this plugin to the kubernetes plugins from backstage, but it makes the finding of an application dynamic. There is a little extra work to be done if using selectors, as we don't use that internally but probably should.

We are running this code internally at this time, but this is the initial approach to get us over the problems we faced when additional clusters were enabled.

Resolves #193

### DIFF
--- a/app-config.yaml
+++ b/app-config.yaml
@@ -136,4 +136,8 @@ catalog:
       rules:
         - allow: [Template]
 argocd:
-  baseUrl: https://localhost:8080
+  # baseUrl: https://localhost:8080
+  username: ${ARGOCD_USERNAME}
+  password: ${ARGOCD_PASSWORD}
+  perCluster:
+    enabled: true
--- a/plugins/backstage-plugin-argo-cd/.eslintrc.js
+++ b/plugins/backstage-plugin-argo-cd/.eslintrc.js
@@ -7,5 +7,6 @@ module.exports = {
     // causing this issue but one of them is. Eventually the @backstage eslint config will
     // be updated to support V3 of @typescript-eslint and this rule can be defaulted.
     '@typescript-eslint/camelcase': 0,
+    'no-console': [0],
   },
 };
--- a/plugins/backstage-plugin-argo-cd/README.md
+++ b/plugins/backstage-plugin-argo-cd/README.md
@@ -6,7 +6,7 @@
 
 ## Features
 
-- 
+-
 
 ## How to add argo-cd project dependency to Backstage app
 
@@ -53,7 +53,7 @@ export { argocdPlugin } from '@roadiehq/backstage-plugin-argo-cd';
 // packages/app/src/components/catalog/EntityPage.tsx
 import {
   EntityArgoCDHistoryCard,
-  isArgocdAvailable
+  isArgocdAvailable,
 } from '@roadiehq/backstage-plugin-argo-cd';
 
 const overviewContent = (
@@ -63,8 +63,8 @@ const overviewContent = (
       <EntitySwitch.Case if={e => Boolean(isArgocdAvailable(e))}>
         <Grid item sm={6}>
           <EntityArgoCDHistoryCard />
-        </Grid> 
-      </EntitySwitch.Case> 
+        </Grid>
+      </EntitySwitch.Case>
     </EntitySwitch>
     ...
   </Grid>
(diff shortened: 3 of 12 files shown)


Xantier:
Thanks for the contribution Karl, 

This looks very promising indeed. Judging from the code it looks like it is close to being backwards compatible/toggleable with the newly added boolean config value so this would work without any config changes for existing usage cases? There are few e2e tests around this plugin that fail to render some of the plugin components but I have a gut feeling that it could be some missing if-clause somewhere to ignore the newly added config values that might fix that.


Coupling the backend functionality with kubernetes-plugin is a bit troublesome and it might be something that we should possibly try to think and refactor away. Are you able to shed some light into what parts of the k8s plugin are used to achieve needed connections etc. to make this work? Would it make sense to lift and shift that functionality into the mentioned 'argocd-backend' plugin instead? Or would it make sense to generalize that functionality from within k8s plugin in the main Backstage repository and share that as a library/separate module that then could be used by 'argocd-backend' (or any other possible plugins that would need such functionality?

Would the backend code in its current state be something that could be added to this repo as a new package/plugin so we could iterate over that one as well?

karlhaworth:
@Xantier - Busy day for me so haven't had a ton of time to look into this. 

I didn't pay close attention to all the tests yet as I wanted input before spending more time on those. It works in our repo, but some things may have changed when I copied it over. This was meant to be a conversation starter :)

The backstage plugin seems to hit all clusters to check for apps based on their [code](https://github.com/backstage/backstage/blob/a1f587c8614adc9d3b81089b28e60595875c7f9e/plugins/kubernetes-backend/src/service/getKubernetesObjectsForServiceHandler.ts#L87). We define a list of clusters as well as a kubernetes backstage selector and it goes to work and find where an app is. What if we did the same? A call to the backend for an app then scans all instances using a backend plugin for an app based on name or selector. That way its not tied to the k8s plugin.

The below is a sample. We use a username and password since the tokens would be different on each instance as stated in my original post. We scan argo instances not tied to clusters. That way we are not hard coding proxy values in catalog-info.yaml files and we're not limited to specific instances? 

```
argocd:
  username: ${ARGOCD_USERNAME}
  password: ${ARGOCD_PASSWORD}
  appLocatorMethods: # defaults to type of proxy if not defined
    - type: 'config'
      instances:
        - name: argo1
          url: https://argo1.test.com
        - name: argo2
          url: https://argo2.test.com
        - name: argo3
          url: https://argo3.test.com
```

karlhaworth:
> Currently the username/password is defined on the top level and that is used across the apps to get a token or connect to directly. Do you think it would be worthwhile to add optional cases where each individual app could have their own username/password blocks as well. The precedence of usage would then be token->individual username/pw -> global username/pw.

I think this would be a good addition and could be something to iterate on later, but at this time I'm thinking it's better to get in workable code, then add that additional scenario. What do you think?

> I'm not certain if that's some kind of setup that is actually in use so feel free to ignore that.

The above is why I say let us hold off on that. 

I will take your suggestion above. I am working on this more today and trying to get it fully wrapped up.

karlhaworth:
#270 was submitted. I'll close this. More discussion can happen on #193 as needed.

PR #217, opened 2021-09-21
Bugsnag plugin

Irma12:
This is an overview table for Bugsnag plugin. Currently it doesn't support project graph and widgets, so that will be added in later iterations

backstage-bugsnag-plugin
 

[Ch4111]


plugins/backstage-plugin-bugsnag/src/components/ErrorGraph/ErrorGraph.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+export const ErrorGraph = ({ bugsnagError }: { bugsnagError: BugsnagError }) => {
+  const api = useApi(bugsnagApiRef);
+  const { value } = useAsync(
+    async () => {
+      return await api.fetchTrends(bugsnagError.project_id, bugsnagError.id);
+    }
+  );
+
+  const data = value?.map(trend => {
```
Xantier, line 33: you could map this directly as the prop to TrendLine component right? `from`/`to` is not used anywhere

plugins/backstage-plugin-bugsnag/src/components/ErrorsOverviewComponent/ErrorsOverview.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const organisationName = useBugsnagData()[0];
+  const projectApiKey = useBugsnagData()[1];
+  const api = useApi(bugsnagApiRef);
+  const [slug, setOrganisationSlug] = useState('');
+
+
+  const { value, loading, error } = useAsync(
+    async () => {
+      const organisations = await api.fetchOrganisations();
+      const organisation = organisations.filter(org => org.name.includes(organisationName))[0];
```
Xantier, line 51: `Array.prototype.find` would achieve the same result here so you don't need array accessors to take the first item. Same thing below

Xantier: You might want to add null/undefined check on the setter as well

plugins/backstage-plugin-bugsnag/src/components/ErrorGraph/ErrorGraph.tsx
Irma12: True, I thought in the future iterations there might be a need for those (when we add filters), so that is the only reason I thought it might be useful to have it there

plugins/backstage-plugin-bugsnag/src/hooks/useBugsnagData.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { useEntity } from "@backstage/plugin-catalog-react";
+
+export const BUGSNAG_ANNOTATIONS = 'bugsnag.com/project-key';
```
Xantier, line 19: should be singular name here IMO

Xantier (dismissed):

plugins/backstage-plugin-bugsnag/README.md
```diff
@@ -0,0 +1,71 @@
+# Bugsnag Plugin for Backstage
+
+![a preview of the Bugsnag plugin](https://raw.githubusercontent.com/RoadieHQ/roadie-backstage-plugins/main/plugins/backstage-plugin-bugsnag/docs/backstage-bugsnag-plugin.png)
```
martina-if, line 3: Nitpick: Can we use a screenshot with the light colors instead?

martina-if:
Very good work. I am happy to ship this although we should work on the tests next as we could start getting contributions and there wouldn't be anything to tell us if someone is breaking.

Xantier (approved):

plugins/backstage-plugin-bugsnag/README.md
Irma12: Sure, I have added it in light colors :)

Irma12:
> Very good work. I am happy to ship this although we should work on the tests next as we could start getting contributions and there wouldn't be anything to tell us if someone is breaking.

@martina-if Thank you, I agree e2e and unit tests are very important to have, so I have updated my PR so it contains basic tests for start

PR #230, opened 2021-10-01
Ch4147 prometheus plugin

Xantier:
Add Prometheus plugin to to Roadie Backstage plugins.

## Components 
1. Graph (line and area, default to line)
2. Alert

Both are exported as standalone as well as as a entity wrapped component so it is usable on entity pages with catalog-info.yaml file config. Additionally includes a 'Content' page which displays both alert table and graphs in a single view. 

Uses proxy to find Prometheus (which is unsecured by default, so likely behind a corporate VPN and/or some reverse proxy). Contacts 2 endpoints, one to get graphs, one to get 'rules' where it can find alerts and their statuses. 

Groups returned graph data based on a dimension user provides, or defaults to first dimension it finds from the returned data. This grouper is used to draw different lines for graphs. Alert endpoint returns all alerts and is not filterable, so displayed data is filtered on Backstage FE. 

### Graphs
`    prometheus.io/rule: memUsage|component,sum by (instance) (node_cpu_seconds_total)`
Comma separated list of query (or pre-saved 'rules') and grouping dimension tuples. Dimension is optional, if omitted first metric name found from the returned data set is used as the key to group items with.
Produces the following graphs:
1. `memUsage|component`
(grouping by component, otherwise `__name__` would be the first item on this saved rule)
image

2. `sum by (instance) (node_cpu_seconds_total)`
 (`instance` is the grouper metric defined in the query --> it is returned on the result set as the first metric name, and is therefore used to group data with.)
image

### Alerts
`    prometheus.io/alert: 'Excessive Memory Usage'`
Comma separated list of alert names configured in Prometheus instance

Produces the following component if 'Excessive Memory Usage' alert is configured on the Prometheus instance, defaults to empty table:
image


packages/entities/test-entity.yaml
```diff
@@ -48,26 +48,12 @@ metadata:
...
-  owner: roadie
-  lifecycle: experimental
-  providesApis:
-    - sample-service-3
-
----
+    prometheus/alert: "test alert name"
+    prometheus/rule: memUsage,node_memory_Active_bytes|instance
 
-apiVersion: backstage.io/v1alpha1
```
Xantier, line 60: Deleting duplicate

plugins/backstage-plugin-prometheus/src/hooks/usePrometheus.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (result.length === 0) {
+    return { keys, metrics, data: [] };
+  }
+
+  const grouper =
+    dimension && result.every(it => Object.keys(it.metric).includes(dimension))
+      ? dimension
+      : Object.keys(result[0].metric)[0];
+
+  const data = _(result)
```
Xantier, line 53: GL someone who tries to fix anything on this chain. Godspeed

dtuite:
Addresses https://github.com/backstage/backstage/issues/714

app-config.yaml
```diff
@@ -67,6 +67,12 @@ proxy:
     headers:
       Cookie: ${ARGOCD_AUTH_TOKEN}
 
+  '/prometheus/api':
+    # url to the api of your hosted argoCD instance
```
julioz, line 71:
```suggestion
    # url to the api of your hosted prometheus instance
```

app-config.yaml
```diff
@@ -67,6 +67,12 @@ proxy:
...
+    target: http://localhost:9090/api/v1/
```
julioz, line 72: is this intended to work with [Thanos](https://thanos.io/tip/components/query.md/) as well?

plugins/backstage-plugin-prometheus/src/api/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      options.configApi.getOptionalString('prometheus.proxyPath') ||
+      DEFAULT_PROXY_PATH;
+  }
+
+  private async getApiUrl() {
+    const proxyUrl = await this.discoveryApi.getBaseUrl('proxy');
+    return `${proxyUrl}${this.proxyPath}`;
+  }
+
+  async query({
```
julioz, line 56: Given the size amount of endpoints this hits, I guess this is ok, but maybe we can consider using https://github.com/siimon/prom-client for the future?

plugins/backstage-plugin-prometheus/src/components/PrometheusAlertStatus/PrometheusAlertEntityWrapper.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { PrometheusAlertEntityWrapper } from './PrometheusAlertEntityWrapper';
+import { prometheusApiRef } from '../../api';
+import { ThemeProvider } from '@material-ui/core';
+import { lightTheme } from '@backstage/theme';
+
+const entityMock = {
+  metadata: {
+    namespace: 'default',
+    annotations: {
+      'prometheus/alert': 'test alert name',
```
julioz, line 33: Following the pattern of other annotations, what about using `prometheus.io` as the prefix (following their domain - https://prometheus.io/)?

plugins/backstage-plugin-prometheus/src/api/index.ts
Xantier: Yeah, we can keep it as is for now. I wasn't aware `prom-client` actually has functionality to fetch data from running prometheus, just to expose metrics.

plugins/backstage-plugin-prometheus/src/components/PrometheusAlertStatus/PrometheusAlertEntityWrapper.test.tsx
Xantier: A good idea yeah :+1:

app-config.yaml
Xantier: If it exposes the same HTTP API as vanilla Prometheus, same queries should work AFAIK.

plugins/backstage-plugin-prometheus/src/api/index.ts
```diff
@@ -0,0 +1,96 @@
...
+    step?: number;
+  }) {
+    const apiUrl = await this.getApiUrl();
+
+    const end = DateTime.now().toSeconds();
+    const start = DateTime.now()
+      .minus(Duration.fromObject(range))
+      .toSeconds();
+    const response = await fetch(
+      `${apiUrl}/query_range?query=${query}&start=${start}&end=${end}&step=${step}`,
```
nicolasarnold12321, line 75: Nice!

Do you think adding validation on the query is overkill? Or should that be controlled by the prom service?

plugins/backstage-plugin-prometheus/src/hooks/usePrometheus.ts
nicolasarnold12321: Ouffft!

plugins/backstage-plugin-prometheus/src/components/PrometheusAlertStatus/PrometheusAlertStatus.tsx
```diff
@@ -0,0 +1,134 @@
...
+const state = (str: string) => {
+  switch (str) {
+    case 'firing':
+      return <StatusError>{str}</StatusError>;
+    case 'pending':
+      return <StatusPending>{str}</StatusPending>;
+    case 'inactive':
+      return <StatusAborted>{str}</StatusAborted>;
+    default:
+      return <StatusAborted>Unknown</StatusAborted>;
```
julioz, line 49: would be nice to include `str` in this case so we have a breadcrumb of what the unknown cases are

nicolasarnold12321 (commented):
Looks great! Just one small question. Do you think it'll be worth adding in an overview card at a later stage?

Maybe displaying the alerts that have fired?

plugins/backstage-plugin-prometheus/src/hooks/usePrometheus.ts
```diff
@@ -0,0 +1,130 @@
...
+  const prometheusApi = useApi(prometheusApiRef);
+  const { value, loading, error } = useAsync(async (): Promise<
+    PrometheusRuleResponse
+  > => {
+    return await prometheusApi.getAlerts();
+  }, []);
+  if (value && value.status !== 'error') {
+    const rules = value.data.groups.flatMap(it => it.rules);
+    const displayableAlerts: PrometheusDisplayableAlert[] = rules
+      .filter(rule => alerts.includes(rule.name))
```
julioz, line 117: is there a possibility an integrator can say "show me all the rules" that doesn't entail copy-pasting each of them in the catalog-info file? for some of our services that would be quite an effort (specially to keep them in sync), the bulkiest ones have dozens of rules.

app-config.yaml
```diff
@@ -67,6 +67,12 @@ proxy:
     headers:
       Cookie: ${ARGOCD_AUTH_TOKEN}
 
+  '/prometheus/api':
+    # url to the api of your hosted prometheus instance
+    target: http://localhost:9090/api/v1/
```
julioz, line 72: another question - can this be somehow overriden through annotations instead in the app-config? internally we have hundreds of prometheus servers deployed, through kubernetes namespaces. that would mean service A and service B send data to different servers (and the metrics are all aggregated by Thanos). assuming Thanos and Prometheus don't have a matching REST API Spec, as an integrator, I would like to be able to specify which path to query for each of my entities (or group of entities, somehow).

how could we model this?

plugins/backstage-plugin-prometheus/src/api/index.ts
Xantier: I think this will be in most cases a copy-paste from a query that is constructed on the prom UI so it's validated in that state already

Xantier:
> Do you think it'll be worth adding in an overview card at a later stage?

Can you tell me a bit more what this would be? There is an overview card for both graph and alerts already that are usable as individual items.

app-config.yaml
julioz: Yes, might need some pointers with the implementation details, but I believe I could, yes.

nicolasarnold12321:
> > Do you think it'll be worth adding in an overview card at a later stage?
> 
> Can you tell me a bit more what this would be? There is an overview card for both graph and alerts already that are usable as individual items.

Ah right I suppose that'll be more than enough so. I was more so thinking that there could be some card that says:
"You have blah alerts firing" which would have a table (or summary) of alerts along with a hyperlink to a component tab.

I think based on what you have said. That is not needed.

.yarnrc
```diff
@@ -0,0 +1 @@
+registry "https://registry.yarnpkg.com"
```
iain-b, line 1: Why this ?

packages/app/package.json
```diff
@@ -35,6 +35,7 @@
     "@roadiehq/backstage-plugin-jira": "^1.0.13",
     "@roadiehq/backstage-plugin-security-insights": "^1.2.11",
     "@roadiehq/backstage-plugin-travis-ci": "^1.0.12",
+    "@roadiehq/plugin-backstage-plugin-prometheus": "^0.1.0",
```
iain-b, line 38: Should be ^1.0.0 ? (the checks pointed this out to me)

iain-b (approved):
Super impressive!

.yarnrc
Xantier: It's for our internal usage so we don't pollute `yarn.lock` with links to our private repo

packages/app/package.json
Xantier: Bumped to be ready for release, didn't remember to update this :+1:

punkle (approved):

PR #237, opened 2021-10-06
Remove trend errors and introduce swr

Irma12:
There is a limit of 10/1 minute for trial and 30/1 standard accounts. However trend will need to make api call for each error and that will cause 429 error appears quite often.
I tried with implementing swr which also could reduces calls and has refresh interval of 60 seconds.
Another option is not using swr at all but also removing trends.


Irma12:
@Xantier I know :( But at this point it just feels as a overhead we don't really have to introduce as it doesn't solve the issue, as you already said. Maybe next time :)

PR #243, opened 2021-10-07
[github-insights] Use scm integration api to look up GitHub host

iain-b:
Use the scm integration api to determine the GitHub location
for an entity rather than read the config directly. The
motivation here is to allow multiple GitHub types to be
used by the same installation e.g. self-hosted and hosted.


plugins/backstage-plugin-github-insights/src/hooks/useEntityGithubScmIntegration.ts
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2020 RoadieHQ
```
Irma12, line 2: Update this notice to the new one :)

plugins/backstage-plugin-github-insights/src/hooks/useRequest.ts
```diff
@@ -28,7 +28,7 @@ export const useRequest = (
   showTotal: boolean = false,
 ) => {
   const auth = useApi(githubAuthApiRef);
-  const { baseUrl } = useUrl();
+  const { baseUrl } = useEntityGithubScmIntegration();
```
Xantier, line 31: (Can't add a comment on the exact line)
Should we add the baseurl as part of the dependency array to the useAsync hook now if the 'loading' state exists when it is triggered. It would default to empty string until loading is done and these calls are done with `[]` deps array.

plugins/backstage-plugin-github-insights/src/components/InsightsPage/InsightsPage.test.tsx
```diff
@@ -13,88 +13,69 @@
...
-  ],
-
-  getOptionalConfig: (_: string) => {
-    return supportConfig;
-  },
-};
-
 const apis = ApiRegistry.from([
   [githubAuthApiRef, new GithubAuth({ getSession } as any)],
-  [configApiRef, config],
```
iain-b, line 66: The test setup changes as we no longer need to mock the config api but we do the scm api

plugins/backstage-plugin-github-insights/src/components/InsightsPage/InsightsPage.test.tsx
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 RoadieHQ
```
iain-b, line 2: Since we had turned off the license check we hadn't been forced to uptake the new one

plugins/backstage-plugin-github-insights/src/components/InsightsPage/InsightsPage.test.tsx
```diff
@@ -13,88 +13,69 @@
...
-                },
-              },
-            }}>
-              <InsightsPage />
-            </EntityProvider>
-          </ThemeProvider>
-        </ApiProvider>
-
-      ));
+      renderResult = render(
```
iain-b, line 51: Just formatting (most files I touched didn't conform to the  current prettier rules)

PR #247, opened 2021-10-08
Wrap Prometheus chip text to multiline

julioz:
With very long values for the chip texts, the text overflows the window width. Here we apply https://github.com/mui-org/material-ui/issues/15185#issuecomment-894307707 to break lines and make the table fit the viewport width.

### Before

Screenshot 2021-10-08 at 18 00 23

### After
Screenshot 2021-10-08 at 17 54 55

Screenshot 2021-10-08 at 17 55 40

### DIFF
--- a/plugins/backstage-plugin-prometheus/package.json
+++ b/plugins/backstage-plugin-prometheus/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@roadiehq/backstage-plugin-prometheus",
-  "version": "1.0.5",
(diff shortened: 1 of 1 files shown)


Irma12:
@julioz Thank you for checking this out and making an improvement. It looks much better now. Could I please ask you just to bump package version also before we merge this.

Irma12:
Thank you for this contribution. It looks great @julioz :)

PR #253, opened 2021-10-20
Add names to all extensions

iamEAP:
Heyo!  👋 

Just a PR that adds names to all of the extensions in all of the plugins (a result of running `npx @backstage/codemods apply extension-names` [from here](https://github.com/backstage/backstage/blob/master/packages/codemods/CHANGELOG.md#0117)), which will be useful to any [analytics instrumentation](https://backstage.io/docs/plugins/analytics) fired in these plugins.


Xantier:
Excellent, thanks @iamEAP!

Could you maybe do a patch version bump to the packages with this PR and we'll ship this

iamEAP:
You got it!

PR #275, opened 2021-11-08
Update readme

Irma12:


README.md
```diff
@@ -78,14 +78,15 @@ yarn dev
 
 ## Structure of the repository.
 
-This repository is a place where all of the RoadieHQ plugins we are developed are integrated under `/plugins` folder. They may be uses and/or modified by following steps below:
+This repository is a place where all of the RoadieHQ plugins we are developed are integrated under `/plugins` folder. Depending on the type of the plugin they are separated in frontend or backend folder. Please note the scaffolder actions is handled separately. Plugins may be used and/or modified by following steps below:
```
iain-b, line 81: is -> are?

README.md
```diff
@@ -78,14 +78,15 @@ yarn dev
...
  
 ### Plugins container
 
 Navigate to 
 
 ```bash
 cd roadie-backstage-plugin/plugins
+cd backend/frontend
```
iain-b, line 89: maybe in this case use the word "or" rather than the "/" as technically that could be a file path and or put <> around it (just to be clear they can't copy paste they have to edit.

iain-b (approved):
One or two tiny things. I'll approve though as we'll have spotty availability

README.md
Irma12: Ah, good catch. Thanks

Issue #71, opened 2021-07-12
New releases should notify Discord and Twitter


dtuite:
Currently, users of our plugins don't really have a way to know when new releases are availalbe.

Can we send an automated discord message to our Discord server when we release a new version? This should work just like the one for Backstage.

Can we send a Tweet from the Roadie Twitter account when we release a new version?

[ch3458]

martina-if:
https://app.clubhouse.io/larder/story/3458/new-releases-should-notify-discord-and-twitter

martina-if:
https://app.clubhouse.io/larder/story/3458/new-releases-should-notify-discord-and-twitter

martina-if:
[Ch3458]

martina-if:
https://app.clubhouse.io/larder/story/3458

PR #280, opened 2021-11-15
Introduce changesets

Irma12:
This PR contains scripts and workflows for changeset changes and automatic releases. I have added Discord webhook and new channel where users should be notified about releases.

I used same scripts in determining tags and doing automatic releases as used in Backstage.
After committing changesets, there should be automatic PR called Version package (as seen here https://github.com/backstage/backstage/pull/7823) and upon merging that, the release job should be triggered.

This PR is probably candidate for pair review


martina-if:
related to #71

Xantier (approved):

Issue #277, opened 2021-11-11
Is there a reason JIRA plugin must assume Basic Auth for activities? 


madorb:
I'm able to successfully pull back issues with:

```
  '/jira/api':
    target: https://my-jira-instance.com
    headers:
      Authorization: 'Bearer <my-token>'
```

but then when it attempts to get activities: it appears to force the use of basic auth, which, of course, doesn't work as we don't have a basic auth value to send. I tried with basic auth, but could not get it to work at all, just got 401s.

https://github.com/RoadieHQ/roadie-backstage-plugins/blob/6d7c927791118d0bb980207f3dfa4188de9ddfef/plugins/frontend/backstage-plugin-jira/src/api/index.ts#L179

If not, it would be better to clarify in the documentation that basic auth is a hard requirement

Irma12:
Hey @madorb thank you for bringing this up. Does it work when you omit '&os_authType=basic` in the request? If so, maybe you could set up a PR so we can take a look? :)

madorb:
it does seem to work when i access it via:
`http://localhost:7000/api/proxy/jira/api/activity?maxResults=25&streams=key+IS+MYPROJECT`
and gets a 401 with:
`http://localhost:7000/api/proxy/jira/api/activity?maxResults=25&streams=key+IS+BP&os_authType=basic`

I suppose you'd need to look at the `Authorization` type in the configuration and if it includes `Bearer` then exclude the attribute, else include it.

Irma12:
Hey @madorb sorry for silence from our side regarding this. 

Using Basic auth is required to retrieve activity in case you want to view restricted content (https://developer.atlassian.com/server/framework/atlassian-sdk/consuming-an-activity-streams-feed/#authentication) but we added the ability to go with Bearer token. This is included in new release (🎉 1.3.0). 

Please, keep in mind you need to add 'jira/token-type: Bearer' annotation to specify if you are using Bearer token, as included in Readme files for the plugin :)

PR #296, opened 2021-11-29
Make bearer token not required in jira plugin

Irma12:
Currently we suppose users will be using Basic auth in order to authenticate. However, some users are using Bearer token so we need to take that into consideration.


Xantier (approved):
Could use a README.md change also.

Irma12:
@Xantier Agree, will add it

martina-if:
fixes #277

.changeset/real-steaks-destroy.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-jira': minor
+---
+
+Provided an option to use Bearer token fo retrieve activity stream.
```
martina-if, line 5: small typo `fo` -> `to`

Xantier (approved):

Issue #266, opened 2021-10-27
[github-plugin-insights] Wrong annotation used to pick github instance


szubster:
Pull request https://github.com/RoadieHQ/roadie-backstage-plugins/pull/243 introduced possibility to attach to multiple github instances.
This is cool, but assumes that yaml file is stored in same place as source code, which is not always the case. Especially when entities are managed not by yaml files, but taken from external systems.

## Expected Behavior

plugin-github-insights uses source location from `backstage.io/source-location` annotation

## Current Behavior

plugin-github-insigths uses annotation `backstage.io/managed-by-location` which in our case points to custom external system and default to github.com

## Steps to Reproduce

Write custom entity provider which sets managed-by-location to url in external system and source-location to github

## Possible Solution

There is dedicated annotation `backstage.io/source-location` and method `getEntitySourceLocation` in package `@backstage/plugin-catalog-react` that can be used.

## Context

We lost support for our ghe instance

## Your Environment

- `@backstage/catalog-react` version: 0.6.0
- `@backstage/integrations` version: 0.6.8

Irma12:
Hey @szubster sorry for the late response, the team was focused on some earlier prioritised items, however we will be discussing this in our next planning sessions and hopefully work on it quite soon.

Looking at the code, I think the main changes need to be done within `plugins/backstage-plugin-github-insights/src/hooks/useEntityGithubScmIntegration.ts` file so we have a check for existence and later usage of  `backstage.io/source-location` instead of automatically taking `backstage.io/managed-by-location` annotations. However, we still haven't started working on it so it may be required in more files.

If this is blocking you I would encourage you to set up a PR as we love contributions and it may be faster way to get the required feature in the production. :)

Irma12:
@szubster This is fixed now and can be seen in new version (v1.3.3) 🎉

PR #314, opened 2021-12-15
Introduce source-location annotations

Irma12:


iain-b (dismissed):
This should fix #266 right ?

plugins/frontend/backstage-plugin-github-insights/src/mocks/mocks.ts
```diff
@@ -18,6 +18,8 @@ export const entityMock = {
     namespace: 'default',
     annotations: {
       'backstage.io/managed-by-location':
+      'url:https://github.com/mcalus3/sample-service/blob/master/backstage5.yaml',
```
iain-b, line 21: That's a real username, perhaps remove it ? (I know it was there before)

Irma12: That is a good point, maybe we should consider making changes in all of the plugins containing it

Irma12:
@iain-b Hopefully yes :)

plugins/frontend/backstage-plugin-github-insights/src/mocks/mocks.ts
iain-b: No harm but no rush or urgency on it though.

PR #333, opened 2021-12-22
Increase scaffolder version and remove frozen flag

Irma12:


.github/workflows/publish.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       uses: actions/setup-node@main
       with:
         node-version: ${{ matrix.node-version }}
-    - run: yarn install --frozen-lockfile
```
Xantier, line 31: I think these should stay in. We want a predetermined dependency versions when we are publishing.

Xantier (approved):

Issue #215, opened 2021-09-17
[github-plugin-insights] Enable multiple integrations 


Irma12:
Currently, plugin will only look at the first integration for Github into cosideration. This is a problem for users having both GH and Github Enterprise integration provided. 

Proposed solution https://discord.com/channels/760879902391009330/760879902391009333/888281121413029888

alexcurtin:
Any updates here?

manusant:
I've been taking a look into the implementation and i believe that the perfect solution should be resolve the github api url form the github integration by matching the host config with the actual **entity source location**.

The source location (location from where the entity was retrieved) should be extracted from entity managed annotations **backstage.io/managed-by-origin-location** or **backstage.io/managed-by-location**.

More documentation on wellknown annotations can be found [here](https://backstage.io/docs/features/software-catalog/well-known-annotations).

@Irma12 please assign the issue to me. I'll submit a PR :)

manusant:
PR opened #338. @Irma12 please have a look. and provide feedback. :)

Irma12:
Closing this as a follow up of https://github.com/RoadieHQ/roadie-backstage-plugins/pull/338

PR #338, opened 2021-12-27
#215:  Enable-multiple-GitHub-integrations-by-extracting-source-location-annotation

manusant:
Signed-off-by: manusant <ney.br.santos@gmail.com>

Enable multiple GitHub integrations by extracting source location from entity managed annotations

### DIFF
--- a/.changeset/ninety-beers-smile.md
+++ b/.changeset/ninety-beers-smile.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-security-insights': patch
+---
+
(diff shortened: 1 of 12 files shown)


Xantier:
Thanks @manusant!

That looks good to me. If CI passes, I think we should be ready to get this merged in.

Xantier (approved):
:shipit:

alexcurtin:
@manusant This is great thanks! How would one use this version of the plugin if we have 1 onprem and 1 github.com - meaning, after using this version and clicking on code insights, the plugin appears to load auth.github from the app-config.yaml which we don't use - we rely on integrations.github with multiple entries

Issue #355, opened 2022-01-03
ArgoCD plugin fails with 404 when fetching app info


cooperbenson-qz:
## Expected Behavior

The request doesn't 404 and I'm able to see the sync information

## Current Behavior

The request to ArgoCD returns a 404 and the ArgoCD card displays this
![image](https://user-images.githubusercontent.com/58954013/147970379-36839378-65fb-40ba-bb1e-7c4bba40f8cd.png)

## Steps to Reproduce

This happens to me if I upgrade to any version of the plugin > 0.13.2

## Possible Solution

Looking at the code and request logs, it's attempting to `GET /api/v1/applications/name/{app name}`, but according to the ArgoCD API documentation that doesn't exist. It's instead expecting `GET /api/v1/applications/{app name}`

## Context

I'm attempting to upgrade to v0.15.2 so I can upgrade other Backstage components, but the upgrade breaks the integration.

## Your Environment

- `@backstage/plugin-catalog-react` version: v0.6.9
- `@backstage/integration` version: v0.7.0
- ArgoCD server version: v2.2.2

sam-robson:
Hey @cooperbenson-qz,

I think I might be responsible for this one! I'm working on a [PR](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/356) which will hopefully fix the issue.

Xantier:
Closing with #356. 
Release version 1.5.3 should have this fixed.

PR #356, opened 2022-01-04
fix: ignore null items from Argo and revert to correct url for the appName single instance pattern

sam-robson:
The card was exploding when the table was trying to load for an app not present in one of the configured instances, while using the argocd-backend-plugin & app-selector pattern. Filtering out null item responses from instances fixes this.

I also added a fix for [issue 355](https://github.com/RoadieHQ/roadie-backstage-plugins/issues/355), by removing the `name` part of the URL for the single-instance/appName pattern.


Xantier (approved):
Thanks @sam-robson!

Merging...

PR #367, opened 2022-01-09
Fix readme image of Bugsnag plugin

dtuite:
Fixes a broken image on the Bugsnag plugin readme.


dtuite:
Closing. We need to take a larger look at readme quality. I opened a ticket instead.

PR #379, opened 2022-01-13
Feature/update search filter to be like GitHub in Pull Requests plugins

fpernin:
We want to have the search bar like in github when we look for an author in a PR or if it is in draft or not with the same format and it happens that with the endpoint of the api search/issues we can access this type of data. There is less data on the PR compared to the old endpoint but it's not a problem because it has all the information we need and it allows a better flexibility in the search parameters.

github-pull-request


Irma12:
@fpernin Thank you for looking into and improving this feature. I think search is very good area to introduce improvements :) 

I am thinking of some use cases and first thing crossing my mind is, currently we have tabs for 'all', 'closed' and 'open' issues. How would this modification affect 'ALL' tab and would it make those tabs completely unneeded? Also, I noticed some strange behaviour after typing few characters and deleting them (blank table). Could you check on your side how does this work for you?
What could be the example of defaultFilter set in annotation? Btw, I think we will need to update README in order to explain new annotations and changes in the way search will function now.

I think we are heading in a great direction but would like us to think different aspects through though.

Irma12:
@fpernin Hey Fred, we are super excited to get this merged so wanted to ask about the progress on resolving the bug found during the testing? Can we help out with anything?

fpernin:
Hey ! 
I corrected the bugs and made some small fixes to make everything work properly.

Irma12:
@fpernin Thank you so much for contributing this feature, I did sanity check and everything seems to be working as expected 🎉 

 I see workflow failing as 'lodash' needs to be added to dependencies, since it causes lint fail. We use changesets for publishing new packages so if you add it we could publish the upgraded version right away  😄 (https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/CONTRIBUTING.md#when-to-use-a-changeset)

Other than that I am happy to approve and merge it 🙌🏼

Irma12:
@fpernin Hey, I updated your branch and I think we are ready to merge it. Thanks for working on this, it looks great ⭐  🚢

fpernin:
Hey, sorry I was overwhelmed thanks for adding lodash for me! Also I'm glad you like the feature!

PR #381, opened 2022-01-17
Sc 5454/bugsnag plugin should fetch all projects

Irma12:
There have been some limitations due the fact Bugsnag retruns only 30 results per page (by default).
Currently Bugsnag supports query by project name so I figured out having 100 items (which I believe is maximum for per_request optional parameter) is enough, in addition of having project name annotation which narrow search a bit more.

### DIFF
--- a/.changeset/honest-poets-behave.md
+++ b/.changeset/honest-poets-behave.md
@@ -1,5 +1,5 @@
 ---
-'@roadiehq/backstage-plugin-bugsnag': patch
+'@roadiehq/backstage-plugin-bugsnag': minor
 ---
 
 Introduce new annotations in order to fix the limit with pagination and default of fetching 30 requests per page.
--- a/.changeset/honest-poets-behave.md
+++ b/.changeset/honest-poets-behave.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-bugsnag': patch
+---
+
+Introduce new annotations in order to fix the limit with pagination and default of fetching 30 requests per page.
--- a/plugins/frontend/backstage-plugin-bugsnag/config.d.ts
+++ b/plugins/frontend/backstage-plugin-bugsnag/config.d.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 Larder Software Limited
+ * Copyright 2022 Larder Software Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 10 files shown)


Xantier:
I think we should just implement the pagination on these tbh. You should be able to use the `base` param to query older items than the last entry on the returned list and implement pagination that way.

Irma12:
@Xantier Yeah I agree that is cleaner approach. However, in https://bugsnagapiv2.docs.apiary.io/#introduction/pagination seems like there is a specific logic for pagination in Bugsnag which made me think there are certain limitations, so I moved from that option. It would be worth reconsidering fetching all the projects and enabling search by name then, but I remember we had API rate limit issues and Bugsnag wasn't very felxible with that 🤔

.changeset/honest-poets-behave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-bugsnag': patch
```
Xantier, line 2: Should we make this minor instead. Would the new annotation option count as a new feature? I think it would

plugins/frontend/backstage-plugin-bugsnag/config.d.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+export interface Config {
+  /** Results per page for bugsnag plugin */
+  bugsnag?: {
+    /**
+     *
+     * @visibility frontend
+     */
+    resultsPerPage: number;
```
Xantier, line 24: Does this need to be optional as well or is it enough if `bugsnag` above is opt?

plugins/frontend/backstage-plugin-bugsnag/src/api/BugsnagClient.ts
```diff
@@ -49,9 +49,14 @@ export class BugsnagClient implements BugsnagApi {
     return payload;
   }
 
-  async fetchProjects(organisationId: string): Promise<Project[]> {
+  async fetchProjects(
+    organisationId?: string,
+    projectName?: string,
+    perPage?: number
```
Xantier, line 55: default value here?

Xantier (approved):
Few items, other than those LGTM

.changeset/honest-poets-behave.md
Irma12: I was thinking about it too, but in the end it seemed as an optimisation and fixing non working stuff rather than new feature so I went with the patch

plugins/frontend/backstage-plugin-bugsnag/config.d.ts
Irma12: Yeah, I think we also need to make it optional, at least by the pattern I see in other schemas

PR #382, opened 2022-01-18
Move React dependencies to peerDependencies and allow both v16 and v17

Rugvip:
Looking to get backstage/backstage#6910 through, and this is the missing piece :grin:

Also moved over one last `@backstage/core-app-api` dependency to devDeps 🧹 


kissmikijr (approved):
Thanks for the updates, looks good!

Issue #387, opened 2022-01-20
[backstage-plugin-jira] Support custom jql queries


tbourrelovh:
We would like to be able to retrieve the number of issues matching a custom JQL query.

## Feature Suggestion

It would be great to be able to provide JQL queries in our entities, alongside the plugin configuration. Those queries would be used to display the number of issues matching them.

## Possible Implementation

I took the liberty to try out a potential implementation, you can see it on my fork : https://github.com/tbourrelovh/roadie-backstage-plugins/commit/12909b06de149e7d9a39881fb6a2027c58227b4b

### Entity configuration

```yaml
metadata:
  ...
  annotations:
    jira/project-key: MYPROJECT
  jira:
    count:
      - title: "P1 & P2"
        query: 'project = MYPROJECT AND issuetype = Incident AND status = Open AND "Application Services DTR" = SERVICE AND Severity in (P1, P2)'
      - title: "P3, P4, P5"
        query: 'project = MYPROJECT AND issuetype = Incident AND status = Open AND "Application Services DTR" = SERVICE AND Severity in (P3, P4, P5)'
      - title: "All bugs"
        query: 'project = MYPROJECT AND issuetype = Bug  AND "Application Services DTR" = SERVICE'
    ...
```

### Display

In term of displaying the data, I thought that it could be added in an another section under the actual count section.

![backstage-plugin-jira custom jql](https://user-images.githubusercontent.com/73471940/150326304-6d06dae2-0fb1-466e-b70a-eec436e75747.png)

## Context

We want to track the number of issues based on custom fields in our jira issues.

kissmikijr:
@tbourrelovh Thanks for the feature suggestion! I think the idea is good, do you mind opening a PR and we could take it to the implementation and get started reviewing it, if you are comfortable with it!:)

tbourrelovh:
Great news  @kissmikijr ! 
I'll gladly do it 👍

tbourrelovh:
I opened the PR https://github.com/RoadieHQ/roadie-backstage-plugins/pull/390

kissmikijr:
Hey @tbourrelovh we have [discussed](https://discord.com/channels/687207715902193673/687235481154617364/940247067736047657) with the maintainers of backstage. They said it is valid however they suggested for this particular use case a backend plugin would be more suitable and I agree, I think it will be better as a backend plugin.

[...]

PR #390, opened 2022-01-21
feat(backstage-plugin-jira): support custom jql queries

tbourrelovh:
Signed-off-by: Thomas Bourrely <thomas.bourrely@ovhcloud.com>

### DIFF
--- a/plugins/frontend/backstage-plugin-jira/src/api/index.ts
+++ b/plugins/frontend/backstage-plugin-jira/src/api/index.ts
@@ -15,7 +15,13 @@
  */
 
 import { createApiRef, DiscoveryApi } from '@backstage/core-plugin-api';
-import { IssuesCounter, IssueType, Project, Status } from '../types';
+import {
+  CustomQuery,
+  IssuesCounter,
+  IssueType,
+  Project,
+  Status,
+} from '../types';
 import fetch from 'cross-fetch';
 
 export const jiraApiRef = createApiRef<JiraAPI>({
@@ -138,7 +144,7 @@ export class JiraAPI {
       iconUrl: status.iconUrl,
     }));
 
-    const filteredIssues = issuesTypes.filter(el => el.name !== "Sub-task");
+    const filteredIssues = issuesTypes.filter(el => el.name !== 'Sub-task');
 
     const issuesCounter = await Promise.all(
       filteredIssues.map(issue => {
@@ -171,11 +177,17 @@ export class JiraAPI {
     };
   }
 
-  async getActivityStream(size: number, projectKey: string, isBearerAuth: boolean) {
+  async getActivityStream(
+    size: number,
+    projectKey: string,
+    isBearerAuth: boolean,
+  ) {
     const { baseUrl } = await this.getUrls();
 
     const request = await fetch(
-      isBearerAuth? `${baseUrl}/activity?maxResults=${size}&streams=key+IS+${projectKey}`:`${baseUrl}/activity?maxResults=${size}&streams=key+IS+${projectKey}&os_authType=basic` ,
+      isBearerAuth
+        ? `${baseUrl}/activity?maxResults=${size}&streams=key+IS+${projectKey}`
+        : `${baseUrl}/activity?maxResults=${size}&streams=key+IS+${projectKey}&os_authType=basic`,
     );
     if (!request.ok) {
       throw new Error(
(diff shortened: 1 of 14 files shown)


plugins/frontend/backstage-plugin-jira/src/api/index.ts
```diff
@@ -217,4 +229,33 @@ export class JiraAPI {
...
+      throw new Error(
+        `failed to fetch data, sttatus : ${request.status}: ${request.statusText}`,
+      );
+    }
+
+    const response = await request.json();
+    return {
+      total: response.total,
+      name: query.title,
+      iconUrl: '', // TODO : find an icon
```
tbourrelovh, line 258: In my implementation I used another component to render the status (CustomQueryStatus) which doesn't require an iconUrl as it uses one from material-ui.

If using this custom status component is kept, maybe the creation of a new type without the `iconUrl` property should be done. Another possibility could be to make this property optional in the already existing type.

plugins/frontend/backstage-plugin-jira/src/api/index.ts
```diff
@@ -217,4 +229,33 @@ export class JiraAPI {
       ),
     ];
   }
+
+  async getIssuesCountByJql(query: CustomQuery) {
+    const { apiUrl } = await this.getUrls();
+
+    // TODO : mutualize this ?
+    const request = await fetch(`${apiUrl}search`, {
```
tbourrelovh, line 237: this is also used in the `getIssuesCountByType ` function, maybe we should create a dedicated function ?

plugins/frontend/backstage-plugin-jira/src/api/index.ts
```diff
@@ -217,4 +229,33 @@ export class JiraAPI {
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        jql: query.query,
```
tbourrelovh, line 243: A co-worker of mine suggested that we filter out "PROJECT = XXXX" from the JQL queries, to avoid requesting data from another project than the one specified in the entity annotation ` jira/project-key: <example-jira-project-key>`

blaryjp: Yes, I think that we should append `PROJECT=${projectFromAnnotation} AND` for each query; and add a check on `query` value to throw an error if we found a `PROJECT=` into it.

kissmikijr:
Thanks, I've started to go through it, one thing that now I understand that this change would mean we would differ from the base backstage schema, since we are expecting/ can act on our custom metadata fields. I do not understand this 100% yet ig there are other possible solution to cover this without introducing the `metadata.jira` keys on the component entity.

Xantier:
Ref https://github.com/RoadieHQ/roadie-backstage-plugins/issues/387

kissmikijr:
@tbourrelovh I'll contact the usptream backstage folks to get some insights about this if it makes sense to put these into the metadata.

tbourrelovh:
Hey @kissmikijr,

I did this after reading https://backstage.io/docs/features/software-catalog/descriptor-format#common-to-all-kinds-the-metadata. Where the following is said : 

```
In addition to these, you may add any number of other fields directly under metadata, 
but be aware that general plugins and tools may not be able to understand their semantics
```

There is another part of the doc about this : https://backstage.io/docs/features/software-catalog/extending-the-model#adding-new-fields-to-the-metadata-object

The other way it could be done is by using annotations like : `jira-custom-query-title-part: query`, but i'm not really a fan of parsing the annotation key to extract the query title.

I'll wait for you to get an answer from the backstage team, there may be some other ways I'm not aware of 🙂

kissmikijr:
Yea It can be done, I am not 100% sure what can be a con in this case. My question is should we do it here and commit for a schema extension.

planeiii-te:
@tbourrelovh There are a lot of unrelated style changes that make the PR a bit tougher to grok. Optimally you want PRs to be as small and focused as possible.

fearphage:
@tbourrelovh There are a lot of unrelated style-only changes that make the PR a bit tougher to grok. Optimally you want PRs to be as small and focused as possible.

testn:
any updates?

punkle:
@tbourrelovh We are happy to close this. Is there an issue for this so its not forgotten?

tbourrelovh:
@punkle It is indeed linked to #387 , I should've linked it earlier 🙂

punkle:
@tbourrelovh feel free to close this pull request if you like. :) we will be happy to review it again when you get the time.

PR #395, opened 2022-01-25
Make check for repo access optional

Irma12:
`getAccessToken` method from @backstage/core-plugin-api' verifies scope of the rights for the repo. However if rejected, this will block fetching even for public repositories. I think we need to decide if there is a need to check for scope in first place, and if not we don't have to show login screen for Github

I introduced new hook (useGithubRepository) in both GH requests and GH insights plugins and idea around it would be to decide on the repo privacy based on data from fetching the repo there.

### DIFF
--- a/plugins/frontend/backstage-plugin-github-insights/.eslintrc.js
+++ b/plugins/frontend/backstage-plugin-github-insights/.eslintrc.js
@@ -1,3 +1,6 @@
 module.exports = {
   extends: [require.resolve('@backstage/cli/config/eslint')],
+  rules: {
+    'notice/notice': 'off',
+  },
 };
--- a/plugins/frontend/backstage-plugin-github-insights/src/hooks/useComplianceHooks.ts
+++ b/plugins/frontend/backstage-plugin-github-insights/src/hooks/useComplianceHooks.ts
@@ -21,13 +21,20 @@ import { useAsync } from 'react-use';
 import { useProjectEntity } from './useProjectEntity';
 import { useEntityGithubScmIntegration } from './useEntityGithubScmIntegration';
 
-export const useProtectedBranches = (entity: Entity) => {
+export const useProtectedBranches = ({
+  entity,
+  isPrivate = false,
+}: {
+  entity: Entity;
+  isPrivate?: boolean;
+}) => {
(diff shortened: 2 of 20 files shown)


plugins/frontend/backstage-plugin-github-insights/.eslintrc.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
   extends: [require.resolve('@backstage/cli/config/eslint')],
+  rules: {
+    'notice/notice': 'off',
```
iain-b, line 4: why off ? this is public, we want this no?

plugins/frontend/backstage-plugin-github-insights/src/hooks/useComplianceHooks.ts
```diff
@@ -48,13 +55,20 @@ export const useProtectedBranches = (entity: Entity) => {
...
+  isPrivate?: boolean;
+}) => {
   const auth = useApi(githubAuthApiRef);
   const { baseUrl } = useEntityGithubScmIntegration(entity);
   const { owner, repo } = useProjectEntity(entity);
   const { value, loading, error } = useAsync(async (): Promise<any> => {
-    const token = await auth.getAccessToken(['repo']);
-    const octokit = new Octokit({ auth: token });
+    const octokit = new Octokit({
+      auth: isPrivate ? await auth.getAccessToken(['repo']) : undefined,
```
iain-b, line 70: Can we check if the user is logged in here? We don't want them to access anonymously if they don't need to (rate limit)

Irma12: Ah, good point 👍🏼

Irma12:
Another approach we can take is in another PR https://github.com/RoadieHQ/roadie-backstage-plugins/pull/397. That would mean we completely remove login screen for these plugins

plugins/frontend/backstage-plugin-github-insights/.eslintrc.js
Irma12: I think we set it like this in all of the plugins, we use eslintrc in root. This worked strange as it reported errors even when notice was added

PR #396, opened 2022-01-25
[SC-5815] polish readmes

kissmikijr:
Restructure and fix the plugin readmes to build on the same schema and fix broken stuff in it

- Fix broken images
- Fix features list where it was missing
- Remove migration notices
- Fix discord invite link, created one that does not expire
- Remove further instructions links


dtuite (dismissed):
I didn't go through each and every README to review, but this certainly looks like an improvement to me.

Irma12:
Make sure to include changesets :) @kissmikijr

kissmikijr:
@Irma12 I intentionally left out the changeset, we are not changing the libraries, I don't think we should publish a new version. These changes only for the repository itself. wdyt?

Irma12:
> @Irma12 I intentionally left out the changeset, we are not changing the libraries, I don't think we should publish a new version. These changes only for the repository itself. wdyt?

README files are shown on npm, so I think we should definitely publish them

kissmikijr:
Ah, yea in this case we definitely should! I tottaly missed this!:) Let me create one quickly

Irma12 (approved):
Looks 👍 good work :)

PR #397, opened 2022-01-26
Make access token calls optional

Irma12:
Pass optional parameter to getAccessToken method

From documentation ->

   * If this is set to true, the user will not be prompted to log in,
   * and an empty response will be returned if there is no existing session.
   *
   * This can be used to perform a check whether the user is logged in, or if you don’t
   * want to force a user to be logged in, but provide functionality if they already are.
   * 
 This means users won’t get annoying Log-in screen. but error message will need to let them know they have to log in. It is still change in a way plugin works, since it won't look for your sign in but at least it won't be causing additional api hit rates to decide if it is public or not.

### DIFF
--- a/.changeset/nine-bats-divide.md
+++ b/.changeset/nine-bats-divide.md
@@ -1,0 +1,6 @@
+---
+'@roadiehq/backstage-plugin-github-insights': minor
+'@roadiehq/backstage-plugin-github-pull-requests': minor
+---
+
+Make getAccessToken calls optional in order to fetch data for public repositories.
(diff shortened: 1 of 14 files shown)


Xantier (approved):

kissmikijr:
The github app and oauth app works differently, the github app does not store the state after a successful login, it uses a refresh token, currently it seems it is not working as [intended](https://github.com/backstage/backstage/issues/9163) since if you refresh your page the frontend does not get recognize you as one who is already logged in. So if we set it to optional, if a plugin user uses a github app integration instead of an oauth app it would mean they need to login after every page start/refresh. 
And my question, where can they login to github if we do not trigger the pop for them anymore?

PR #402, opened 2022-01-31
SC-6275 add etag handling to github-insights

kissmikijr:


.changeset/few-frogs-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+'@roadiehq/backstage-plugin-github-insights': major
+'app': minor
```
Xantier, line 3: No need to change app versions

.changeset/few-frogs-tease.md
```diff
@@ -0,0 +1,6 @@
+---
+'@roadiehq/backstage-plugin-github-insights': major
```
Xantier, line 2: Maybe you can wrap the exported components with the new Provider internally and export those already wrapped. That way we wouldn't need to do a breaking change.

kissmikijr: It wouldn't work in our case, because the state lives inside the provider. So we will need to put these providers in a place in the app where it won't get unmounted so it can keep the state when we switch pages/tabs and move around in the app

Xantier: Hmm, that's a pity. I'm not a big fan of adding new Context/Providers to the root of the app since this leaks the configuration needed by the plugin to a higher level. The whole app doesn't care about this state, only a tiny portion of our GitHub plugin does.

Could you possibly experiment with global state without Context to store this data and we can see how that looks. It would need initialization function call as well but it would be just a function call and would wrap the whole application. 

Random articles from the internet:
https://formidable.com/blog/2021/stores-no-context-api/
https://blog.openreplay.com/steps-to-develop-global-state-for-react-with-hooks-without-context

Maybe start a discussion on Discord as well for more input?

kissmikijr: Yea, it is :/ I don't really like it either this way, it feels wrong. I'll check out to use the global state. Other idea without provider we could use localstorage (I think it's still really hakcy). I've been reading about the HTTP 304 and I've read that the browser should use its cache for the request when we get 304, but I was not able to make it work that way initially, I'll check again also this one.

kissmikijr:
[SC-6501]

plugins/frontend/backstage-plugin-github-insights/src/hooks/useComplianceHooks.ts
```diff
@@ -20,27 +20,55 @@ import { OctokitResponse } from '@octokit/types';
...
 
+  useEffect(() => {
+    if (value) {
+      setBranches(value)
+    }
+
+  }, [value, setBranches])
+
   return {
     branches: value,
```
Irma12, line 73: Could we return value.data from here?

plugins/frontend/backstage-plugin-github-insights/src/components/Widgets/ContributorsCard/ContributorsCard.test.tsx
```diff
@@ -81,7 +81,42 @@ describe('ContributorsCard', () => {
       ),
     );
 
-    expect(await rendered.findByText('People')).toBeInTheDocument();
-    expect(await rendered.getByAltText('Rugvip')).toBeInTheDocument();
+    expect(await screen.findByText('People')).toBeInTheDocument();
+    expect(await screen.getByAltText('Rugvip')).toBeInTheDocument();
   });
+  it('should display a card with data from sate on second render when request is 304', async () => {
```
Irma12, line 87: state :)

plugins/frontend/backstage-plugin-github-insights/src/hooks/useContributor.ts
```diff
@@ -18,29 +18,49 @@ import { useAsync } from 'react-use';
...
 import { useEntityGithubScmIntegration } from './useEntityGithubScmIntegration';
-import { ContributorData } from '../components/types';
+import { ContributorData, GithubRequestState } from '../components/types';
 import { useEntity } from '@backstage/plugin-catalog-react';
+import { useStore } from '../components/store';
+import { RequestError } from "@octokit/request-error";
+import { useEffect } from 'react';
 
-export const useContributor = (username: string) => {
+export const useContributor = (username: string): { contributor?: ContributorData, error?: Error, loading: boolean } => {
```
Irma12, line 27: Should we follow the same pattern for other hooks? :)

plugins/frontend/backstage-plugin-github-insights/src/hooks/useComplianceHooks.ts
kissmikijr: Done

plugins/frontend/backstage-plugin-github-insights/src/components/Widgets/ContributorsCard/ContributorsCard.test.tsx
kissmikijr: fixed!

plugins/frontend/backstage-plugin-github-insights/src/components/Widgets/ComplianceCard/ComplianceCard.test.tsx
```diff
@@ -70,14 +70,14 @@ describe('ComplianceCard', () => {
         (_, res, ctx) => res(ctx.json(licenseResponseMock)),
       ),
       rest.get(
-        'https://api.github.com/repos/mcalus3/backstage/branches?protected=true',
+        'https://api.github.com/repos/mcalus3/backstage/branches',
```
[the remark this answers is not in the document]
Xantier: Yup, according to docs these params should be handled in the passed in function like `req.params whatever` instead of in the URL

PR #404, opened 2022-02-03
Gracefully handle situation where an app is missing operationState

### LINKED ISSUE #401: plugin-argocd errors out when operationState is missing from app status
In some situations - when an app has never been synced, I believe - the argo instance's status will simply be missing the application state. In this case, the [type-checking will fail](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/89ecef2e53aa7d3392c7277bb72044ea38c11934/plugins/frontend/backstage-plugin-argo-cd/src/types.ts#L21-L24), and a very long type-error will be printed out in plain-text in the ArgoCD card. 

## Expected Behavior

The UI components should degrade gracefully and simply not show information about the operation state.

## Current Behavior

The components are rendered with the text content:

```
Error occurred while fetching data. Error: remote data validation failed: Expecting one of: Array<{ metadata: { name: string, instance: ({ name: (string | undefined), url: (string | undefined) } | undefined) }, status: { sync: { status: string }, he... Array<{ metadata: { name: string, instance: ({ name: (string | undefined), url: (string | undefined) } | undefined) }, status: { sync: { status: string }, he... null at items but instead got: ( ... here follows the entire HTTP response body ... )

## Steps to Reproduce

I think you should be able to reproduce this by creating an app in ArgoCD but _not_ syncing it.

Then add all the relevant selectors and so on to a Backstage resource. 

## Possible Solution

Allow `undefined` as a value in the type-check. Then make sure the UI code handles its absence down the pipe.

## Context

We think this has caused problems for us as a result of a recent migration of some apps from one ArgoCD instance to another; until the app is actually deployed and argo sync called, the UI is broken for these apps. 

## Your Environment

- `@backstage/plugin-catalog-react` version: 0.6.12
- `@backstage/integration` version: 0.7.2

tragiclifestories:
Fixes #401.

### DIFF
--- a/.changeset/two-melons-smash.md
+++ b/.changeset/two-melons-smash.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-argo-cd': patch
+---
+
+Gracefully handle situation where an app is missing operationState
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDDetailsCard.tsx
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDDetailsCard.tsx
@@ -39,7 +39,7 @@ const getElapsedTime = (start: string) => {
   return moment(start).fromNow();
 };
 
-const State = ({ value }: { value: string; }) => {
+const State = ({ value }: { value: string }) => {
   const colorMap: Record<string, string> = {
     Pending: '#dcbc21',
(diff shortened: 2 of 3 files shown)


Irma12:
Hey @tragiclifestories Thank you so much for fixing this. Everything looks great 👌🏼 🎉

Irma12 (approved):

PR #407, opened 2022-02-07
Fix typo and add additional value in cypress

Irma12:
I noticed we have a typo in cypress tests so fixed it.


Irma12:
I will include this in 'Update dependecies' PR (#408 )

PR #408, opened 2022-02-09
Update dependencies

Irma12:
Due changes in https://github.com/backstage/backstage/releases/tag/v0.66.0 (@backstage/plugin-auth-backend@0.9.0) and several breaking changes I refactored some files.

Because of cypress tests failure I increased timeout and removed some unecessary files


packages/app/cypress/integration/backstage-github-insights.ts
```diff
@@ -25,22 +25,14 @@ describe('GithubInsights', () => {
         cy.intercept('GET', 'https://api.github.com/repos/organisation/github-project-slug/readme', { fixture: 'githubInsights/readme.json' })
         cy.intercept('GET', 'https://api.github.com/repos/organisation/github-project-slug/branches?protected=true', { fixture: 'githubInsights/compliance.json' })
         cy.intercept('GET', 'https://api.github.com/repos/organisation/github-project-slug/contributors?per_page=10', { fixture: 'githubInsights/contributors.json' })
-        cy.visit('/catalog/default/component/sample-service')
     })
 
     describe('Navigating to GitHub Insights', () => {
         it('should show GitHub Insights Releases in Overview tab', () => {
+            cy.visit('/catalog/default/component/sample-service')
```
iain-b, line 32: Out of curiosity why move this from the beforeEach?

Irma12: Well, cy.visit() takes 30000 ms per default before timeout so I was trying to avoid situation of timeout before we actually do the test. But now that I think about maybe we can keep it in beforeEach as I already increased the response timeout. @iain-b

PR #410, opened 2022-02-11
adds a shell:run action using the bash container image

punkle:


punkle:
Here is what it looks like:

image

Xantier:
I think we should map out the security implications of this before merging

PR #425, opened 2022-02-23
[SC-6776] add aws actions package

kissmikijr:


plugins/scaffolder-actions/scaffolder-backend-module-aws/README.md
```diff
@@ -0,0 +1,123 @@
+# scaffolder-backend-module-utils actions package
+
+Welcome to the roadie `aws` actions for the `scaffolder-backend`.
+
+This contains a collection of actions:
```
punkle, line 5: Collection of one "action" :D

plugins/scaffolder-actions/scaffolder-backend-module-aws/README.md
```diff
@@ -0,0 +1,123 @@
...
+  database,
+  catalogClient,
+  reader,
+  actions,
+});
+```
+
+# Authentication
+
+This action accepts an optional `credentials` parameter. Which should be a `CredentialProvider` interface from the [@aws-sdk/credential-provider](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_credential_providers.html) If you want to override the `@aws-sdk/client-s3` module's default authentication, provide a valid credential to the action.
```
punkle, line 59: : ❤️

plugins/scaffolder-actions/scaffolder-backend-module-aws/src/actions/s3/cp.test.ts
```diff
@@ -0,0 +1,173 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { getVoidLogger } from '@backstage/backend-common';
+import { createAwsS3CpAction } from './cp';
+import { PassThrough } from 'stream';
+import mockFs from 'mock-fs';
+
+const mockS3Client = {
```
punkle, line 21: There is a library called aws-sdk-client-mock that I think will be easier to work with.

punkle (approved):

plugins/scaffolder-actions/scaffolder-backend-module-aws/src/actions/s3/cp.test.ts
kissmikijr: Yes, I tried to make it work, however I was not able to There was an ugly `unable to find module something something error` Let me repor again and post it here why I didnt use that. I think it's something with the transpile config or idk. So I opted for now to manually mock it. I suggest we try to use the aws sdk client mock when we add more actions here and it'll need in more packages

plugins/scaffolder-actions/scaffolder-backend-module-aws/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "@roadiehq/scaffolder-backend-module-aws",
+  "version": "0.1.0",
```
Xantier, line 3: Let's start with 1.0.0. It is better for users to know what is actually happening when we do modification and using semver properly is the way to do that.

plugins/scaffolder-actions/scaffolder-backend-module-aws/src/actions/s3/cp.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export function createAwsS3CpAction(options?: {
+  credentials?: CredentialProvider;
+}) {
+  return createTemplateAction<{
+    path?: string;
+    prefix?: string;
+    bucket: string;
+    region: string;
+  }>({
```
Xantier, line 33: We can extend this by adding
```
      endpoint?: string
      s3ForcePathStyle?: boolean
```
to support a ton of "AWS S3 compatible" storages. These can be passed into the config with those names so it's a quick addition.

Can be done in the next iteration if you don't want to tackle it now.

kissmikijr: Added them as you said it was a super easy addition

plugins/scaffolder-actions/scaffolder-backend-module-aws/package.json
kissmikijr: done

Xantier (approved):

PR #431, opened 2022-02-26
@roadiehq/scaffolder-backend-module-http-request: Allow specifying absolute Urls in the Path 

ewassef:
This will allow the callers to specify absolute paths as well as relative ones

- [ x] Added tests for new functionality and regression tests for bug fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Installation instructions for each plugin can be found in their individual READM
 
 Backstage is an open platform for creating developer portals. To learn more about the problems it can help solve, please check out our [Ultimate Guide to Backstage by Spotify](https://roadie.io/backstage-spotify/).
 
-## 
+##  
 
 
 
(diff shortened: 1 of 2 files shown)


ewassef:
@kissmikijr I noticed you were one of the latest committers

punkle:
@Xantier apologies, just seeing your comment now after posting mine.

ewassef:
> Hi @ewassef ,
> 
> Thanks for the contribution! I think this was one of the functionalities that was originally in the action already but we backed out from because it introduces a security hole without much possibilities to mitigate the blast radius. Having the URLs and calls be backed by Backstage will give admins the possibility to use Backstage configs to expose only few known URLs.
> 
> We could possible explore some ways to make this an optional functionality for users that necessary want to run this unsecurely, maybe via some configuration options.

Thank you for the reply. Can you give me some more details on the security issues it exposes? 
We will be attaching a sidecar to the backend written in .net to expose some business functionality. 
We explored creating a backend-plugin that was a simple proxy but this seemed to be a better option.

Are there any possible options that you are referring to that I can implement and contribute?

Xantier:
> Are there any possible options that you are referring to that I can implement and contribute?

You could possibly create a [proxy configuration](https://backstage.io/docs/plugins/proxying#docsNav) in your Backstage instance that proxies the traffic to the sidecar service. It is usually better to do it this way, since Backstage configurations are more strictly guarded and managed by a dedicated team that administers Backstage whereas scaffolder templates etc. could possibly be done by anyone in the company.

ewassef:
@Xantier that makes much more sense! I think this part was missed in our reading of the docs.. ill close this PR and use that instead.

PR #445, opened 2022-03-10
SC-7421 do not force auth

kissmikijr:
Do not force the github auth pop up. It will render the card with a message to log in to github if you'd like to see it.

image

### DIFF
--- a/plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
+++ b/plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
@@ -44,11 +44,11 @@ export type MarkdownContentProps = {
  * @public
  */
 export const Content = (props: MarkdownContentProps) => {
-  const auth = useApi(githubAuthApiRef);
+  const githubApi = useApi(githubAuthApiRef);
   const [isLoggedIn, setIsLoggedIn] = useState(false);
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
```diff
@@ -37,6 +44,28 @@ export type MarkdownContentProps = {
  * @public
  */
 export const Content = (props: MarkdownContentProps) => {
+  const githubApi = useApi(githubAuthApiRef);
```
iain-b, line 47: If you're feeling motivated this could be a separate custom hook (e.g. useGithubLoginState) ? Ideally we'd put it somewhere where it could be shared easily but I don't think we have anywhere... but extracting it might be the first step.

iain-b (approved):
One observation but this is a nice change 👍

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
kissmikijr: Yea, good idea, we have a LOT of github plugins so probably I'll move it out, but in a separate PR, I'd like to merge this to not force the login popup in roadie.

iain-b (approved):

iain-b (approved):

PR #447, opened 2022-03-10
Fix release for home plugins

kissmikijr:
This should fix the release of the home plugins if they are the only ones that are changed in a PR.
Tweaks some PR titles in the update dependencies and version packages jobs.
Updates the markdown plugin's readme to create a new release right away

### DIFF
--- a/.github/workflows/changeset.yml
+++ b/.github/workflows/changeset.yml
@@ -18,5 +18,6 @@ jobs:
         with:
           # Calls out to `changeset version`, but also runs prettier
           version: yarn release
+          title: Version Packages - plugins
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/changeset.yml
```diff
@@ -18,5 +18,6 @@ jobs:
         with:
           # Calls out to `changeset version`, but also runs prettier
           version: yarn release
+          title: Version Packages - plugins
```
kissmikijr, line 21: Want to add the `- plugins` into the PR title so the ticket that will be created will be named the same. It should make easier for us to decide which ticket is for which repo

punkle (approved):

PR #464, opened 2022-03-28
Updated documentation to include conditionals

liam-mackie:
When implementing the Github PR plugin and the Security Insights plugin, we found that the default documentation did not include entity switches on some of the resources, even though it's exposed by the plugin. This PR adds the switches to the example code in the documentation.


kissmikijr:
Thanks for this addition!! 🎉 
One small thing to get it released after merge could you pls add a changeset ? You'll need to run `yarn changeset` in the root.

kissmikijr (approved):

Issue #461, opened 2022-03-22
Add Average Size Of PR to Pull Request Statistics


hyanghm:
In an early development screenshot I found in [one Backstage discussion thread](https://github.com/backstage/backstage/issues/1808#issuecomment-671827496), there was a metric for Average Size of PR in Pull Request Statistics. We do not have it in the current implementation of the plugin. This is a very valuable metric. It helps teams analyze their size of PR's and figure out how to reduce time commitment for code reviews. 

## Feature Suggestion

Add Average Size Of PR to Pull Request Statistics

## Possible Implementation

Add the metric to the pull request statistics widget.

kissmikijr:
Thanks for opening this issue! As we discussed on discord this is a good idea! I'll keep this ticket updated/linked to the PR as we will start progressing on this.

PR #468, opened 2022-03-30
[SC-7839] - add average size of pr

kissmikijr:
image

First iteration for this request: https://github.com/RoadieHQ/roadie-backstage-plugins/issues/461


plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithuRepositoryData.tsx
```diff
@@ -26,14 +26,10 @@ export const useGithuRepositoryData = (url: string) => {
...
-    const response = await new Octokit({
-      auth: token,
-      ...(baseUrl && { baseUrl }),
-    }).request({ url: url });
-
-    return {
-      htmlUrl: response.data.html_url,
-      fullName: response.data.full_name,
-    };
+    return new GithubPullRequestsClient().getRepositoryData({
```
kissmikijr, line 29: I am not sure about this :/ So I was needed this logic from this hook inside an async hook, so I took the logic and put it into the api so it could be used with the useApi() hook. However this hook is used in the homepage componenets, and in the homepage we do not have access to the api provider so actually we are not able to use it via that hook that's why I opted for this solution. 
I am unsure what drawbacks it can have with instantiating a new client on every hook call (I guess)

My other thought is maybe it could work if we setup the homepage a way that it could use the useApi() hook ? 
What do you think ?

Xantier: I think it's ok to construct the client locally for simplicity sake. Could be better to pull the initialization outside of the hook though. You could even possibly wrap that into a useMemo with an empty dep array or something but it's likely more overhead perf wise to do that than initializing it every time. The class itself is tiny so I wouldn't worry too much about it

Xantier (dismissed):

plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithuRepositoryData.tsx
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 import { useApi, githubAuthApiRef } from '@backstage/core-plugin-api';
-import { Octokit } from '@octokit/rest';
 import { useAsync } from 'react-use';
 import { GithubRepositoryData } from '../types';
 import { useBaseUrl } from './useBaseUrl';
+import { GithubPullRequestsClient } from '../api';
 
 export const useGithuRepositoryData = (url: string) => {
```
Irma12, line 22: I know this isn’t changed not but could we fix the typo here and add ‘b’ in here so it is ‘useGithubRepositoryData’ 😁

kissmikijr: 🤦🏿  ofc! thanks for catching it!

Irma12 (approved):

PR #474, opened 2022-04-12
Add Iframe plugin to render iframes in backstage

nicolasarnold12321:
This here provides the @roadiehq/backstage-plugin-iframe plugin.

It supports iframes in backstage and allows you to customize its:
* src
* title
* height
* width

The src field is mandatory.

This is it's definition:
```
export { iframePlugin, EntityIFrameCard, } from './plugin';

type IFrameProps = {
    src: string;
    title?: string | undefined;
    height?: string | undefined;
    width?: string | undefined;
    class?: string | undefined;
}
```

Signed-off-by: Nicolas Arnold <nic@roadie.io>

Screenshot 2022-04-12 at 17 23 03


dtuite:
Do we need to be worried about security? I noticed that TechDocs has an allowlist of iframe hosts which can be embedded. This is [mentioned in our Slack here](https://roadiehq.slack.com/archives/C031N5YKNNA/p1649320010120149). Whatever security concerns drove that consideration would surely apply here too.

Xantier:
This could use a Content and Page type components as well

plugins/frontend/backstage-plugin-iframe/src/plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+        import('./components/IFrameComponent').then(
+          m => m.IFrameCard,
+        ),
+    },
+  }),
+);
+
+export const EntityIFramePage = iframePlugin.provide(
+  createRoutableExtension({
+    name: 'EntityBugsnagErrorsOverviewTable',
```
Xantier, line 50:
```suggestion
    name: 'EntityIFramePage',
```

plugins/frontend/backstage-plugin-iframe/src/plugin.ts
```diff
@@ -0,0 +1,56 @@
...
+    component: {
+      lazy: () =>
+        import('./components/IFrameComponent').then(
+          m => m.IFrameCard,
+        ),
+    },
+  }),
+);
+
+export const EntityIFramePage = iframePlugin.provide(
```
Xantier, line 48: This is using 'content' routeref. Should this be a Content type instead of a Page?

plugins/frontend/backstage-plugin-iframe/README.md
```diff
@@ -0,0 +1,73 @@
...
+    }
+  ]
+}
+const serviceEntityPage = (
+  <EntityLayoutWrapper>
+    ...
+    <EntityLayout.Route 
+      path="/mycustom-iframes"
+      title="Iframes">
+        <EntityIFramePage {...iframeProps} />
```
Xantier, line 39: EntityPages should only have Content type components within them as tabs. 
Take a look here for the conventions: https://backstage.io/docs/plugins/composability#naming-patterns

plugins/frontend/backstage-plugin-iframe/README.md
```diff
@@ -0,0 +1,73 @@
...
+    ...
+  </Grid>
+);
+```
+
+## Allowlisting
+
+This particular plugin supports allowlisting. What this means is you can add a domain to the plugin's configuration that will be verified during the creation of the plugins components.
+
+The config is like so:
```
Irma12, line 64: I would suggest making it a bit clearer where this is configured so maybe just adding a 
// app-config.yaml here or in description above :)

nicolasarnold12321: Thank you Irma! That makes sense

plugins/frontend/backstage-plugin-iframe/README.md
nicolasarnold12321: Ah ok, I see.

Thank you @Xantier

PR #475, opened 2022-04-12
feat: add create and delete to argocd

### LINKED ISSUE #473: Argo CD Create/Delete Endpoints and Scaffolder Action
## Feature Suggestion

Argo CD Create/Delete options for use with create plugin

**Is this something desired by the community?** If the contribution will be accepted, we can discuss open sourcing what we have already built. We would be moving from read-only to read/write permissions, however that can be controlled at the Argo CD level.

## Possible Implementation

American Airlines has create and delete endpoints built into the backend plugin that is open sourced.
We use the delete endpoint in a custom catalog component.
For the create endpoint, we incorporated it into a scaffolder action and the above mentioned custom catalog component.

We are proposing to open source the create and delete endpoints on the backend as well as the scaffolder action for Argo CD create.

## Context

Allowing Argo CD projects and applications to be created from the scaffolding create plugin.
Allowing deletion of projects and applications.

Right now due to code differences we manually pull-in contributed upstream code to our local and we'd prefer to contribute to open source.

karlhaworth:
Closes #473

feat: add create and delete to argocd

Will bring in the below:
- Add create endpoints
- Add delete endpoints
- Add sync endpoints
- Add scaffolder action for create


Xantier:
Thanks Karl!
Took a quick peek at this and it's is looking like a brilliant addition to the plugin. The feature set on this is starting to grow into a very useful tool in itself. 

One thing that came to mind while reading through the code was to notify/advice the user in cases where their Argo token doesn't have the correct permissions to do create-update-delete actions. Could there be a possibility to identify that kind of information and display the actions based on the permissions or alternatively return some helpful error messages when an unauthed action is taken?

karlhaworth:
Thanks @Xantier i agree. I felt like it was still slightly rough in a few areas which is why it's in draft. 
I appreciate you looking through this as draft. It helps.
I'll address that and hopefully make this an active PR by end of week.

karlhaworth:
Below is the scaffold service on the backend throwing an error (with proper unit tests) and displaying a permission issue on the credential being used.

image

`2022-04-13T15:30:47.000Z Error: permission denied: projects, create, khtest20220413-2-nonprod, sub: runwaytestuser18471, iat: 2022-04-13T15:30:46Z`

I knew more on the error handling side which is my focus today. Thank you again.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -1,32 +1,89 @@
...
     baseUrl: string,
     argoInstanceName: string,
     options: {
-      name: string,
+      name: string;
       selector: string;
     },
     argoToken: string,
   ) => Promise<object>;
+  createArgoProject: (
```
Xantier, line 27: Could it be possible to modify these methods to accept an object instead of direct arguments instead. That way we can more easily see on the caller side what string is assigned to which variable. Also gives more possibilities to modify these later since changes to the method signature don't introduce a breaking changes

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -130,4 +208,350 @@ export class ArgoService implements ArgoServiceApi {
...
+      project: {
+        metadata: {
+          name: projectName,
+        },
+        spec: {
+          destinations: [
+            {
+              name: 'local',
+              namespace: namespace,
+              server: 'https://kubernetes.default.svc',
```
Xantier, line 229: Would this be something that may need different addresses for different cluster configurations or is it always the same?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
```diff
@@ -119,6 +121,181 @@ export function createRouter({
...
 
+  router.post('/createArgo', async (request, response) => {
+    const argoInstanceName = request.body.clusterName;
+    const namespace = request.body.namespace;
+    const projectName = request.body.projectName as string;
+    const appName = request.body.appName as string;
+    const labelValue = appName;
+    const sourceRepo = request.body.sourceRepo;
+    const sourcePath = request.body.sourcePath;
+    const argoSvc = new ArgoService(argoUserName, argoPassword, config);
```
Xantier, line 132: Now that we are starting to have more of these endpoints in this router, would it make sense to initialize this `ArgoService` outside of the endpoint itself and reuse the same instance on all of them?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
```diff
@@ -119,6 +121,181 @@ export function createRouter({
...
+      let argoDeleteProjectResp: boolean;
+      try {
+        let argoApp = await argoSvc.getArgoAppData(
+          matchedArgoInstance.url,
+          matchedArgoInstance.name,
+          { name: argoAppName },
+          token,
+        );
+        let isAppDeployed = 'metadata' in argoApp;
+        for (let attempts = 0; attempts < 25 && isAppDeployed; attempts++) {
```
Xantier, line 271: Would this ever take longer than 75 seconds on larger apps? Maybe we could make this an optional config value just in case

Xantier (commented):
This is excellent. Thanks Karl,

Few small nitpicks, but other than those this looks to me like it's good to go

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
karlhaworth: In our case its always the same however I see your point. I have made this an optional input parameter with a default value.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
karlhaworth: 100%. changing

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
karlhaworth: Feels like it shouldn't however I added this as an optional configuration item `waitCycles`

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
karlhaworth: I believe I have done what is requested. Let me know if it works for you or if I missed what you were saying. Thanks.

Xantier (approved):
Excellent. This looks good to me so I will merge this and carve out a release of this functionality.

PR #489, opened 2022-04-21
Hide instance column in argo cards when using a single argo instance

stefanbuck:

- [ ] Added tests for new functionality and regression tests for bug fixes (I'm going to add tests once we agreed on a solution, see comment below)

When using a single Argo instance, the column `Instance` is empty and takes up unnecessary space. In the current implementation I'm using the `argocd.appLocatorMethods` config field as an indicator to toggle the column. Maybe there is a better way of identifying this given there are two options available to make the plugin talk to multiple argo instances. I think this only works in the cases of [option 2](https://github.com/RoadieHQ/roadie-backstage-plugins/tree/main/plugins/frontend/backstage-plugin-argo-cd#support-for-multiple-argo-cd-instances---option-2---argo-cd-backend-plugin) and I wonder if there is a way to do the same in case people opt for [option 1](https://github.com/RoadieHQ/roadie-backstage-plugins/tree/main/plugins/frontend/backstage-plugin-argo-cd#support-for-multiple-argocd-instances---option-1). What are you thoughts?

### Before
Screenshot 2022-04-21 at 09 40 14

### After
Screenshot 2022-04-21 at 09 39 22

### DIFF
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDDetailsCard.tsx
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDDetailsCard.tsx
@@ -79,16 +79,13 @@ const OverviewComponent = ({
 }: OverviewComponentProps) => {
   const configApi = useApi(configApiRef);
   const baseUrl = configApi.getOptionalString('argocd.baseUrl');
+  const supportsMultipleArgoInstances: boolean = Boolean(configApi.getOptionalConfigArray('argocd.appLocatorMethods')?.length);
 
   const columns: TableColumn[] = [
     {
       title: 'Name',
       highlight: true,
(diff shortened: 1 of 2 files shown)


Xantier (approved):
Thanks @stefanbuck!

Could you add a patch changeset to this and we'll merge this in and craft a release.

stefanbuck:
I thought I left a comment asking how to proceed with setups using [option 1](https://github.com/RoadieHQ/roadie-backstage-plugins/tree/main/plugins/frontend/backstage-plugin-argo-cd?rgh-link-date=2022-04-21T07%3A52%3A58Z#support-for-multiple-argocd-instances---option-1), but looks like I did not. 

Anyway, thanks for taking care of the missing changeset

PR #495, opened 2022-04-22
Create shortcut plugin

Irma12:
This plugin will give an option to preview Shortcut stories currently in progress and it can be used separately and included in Homepage Plugin also as a widget.

I have also made small fix for Home Page so it looks a bit neater.
Before:
Screenshot 2022-04-22 at 12 12 48

After:
Screenshot 2022-04-22 at 10 19 32


plugins/frontend/backstage-plugin-shortcut/README.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+3. Add plugin to your Backstage instance:
+
+```ts
+// packages/app/src/components/catalog/EntityPage.tsx
+import { EntityShorcutPageContent } from '@roadiehq/backstage-plugin-shortcut';
+...
+
+const serviceEntityPage = (
```
kissmikijr, line 48: I think this plugin works for a user, not strictly on a per service level. I'd not add this here. I think we should encourage the use of the homepage plugin and /or using a full page that is accessible from the sidebar for it

Irma12: For the encouragement for the  usage of homepage plugin, I would still like to have this option open for OSS community in a way they would use it how it benefits them. That is one of the reasons I decided to go with creating a StoriesCard in ShortcutPage.

That way it can also be used in EntityPage as a widget. However, I do agree regarding service level, so definitely could move it there.

plugins/frontend/backstage-plugin-shortcut/src/components/Widgets/StoriesCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+import Star from '@material-ui/icons/Star';
+import Build from '@material-ui/icons/Build';
+import { shortcutApiRef } from '../../api';
+import { BackstageTheme } from '@backstage/theme';
+
+const useStyles = makeStyles<BackstageTheme>(theme => ({
+  listStyle: {
+    borderRadius: '1rem',
+    boxShadow:
+      '5px 0 5px -5px rgb(0 0 0 / 70%), -5px 0 5px -5px rgb(0 0 0 / 70%)',
```
kissmikijr, line 41: there are boxShadows in the theme. Can you check if there is one that suits this shadow? If not it's ok.
for example:
```
theme.shadows[10]
```

plugins/frontend/backstage-plugin-shortcut/src/components/Widgets/StoriesCard.tsx
```diff
@@ -0,0 +1,237 @@
...
+              {story.started ? 'In progress' : 'Not started'}
+            </Typography>
+          </Grid>
+        </Grid>
+      </Grid>
+    </Box>
+  );
+};
+
+const StoriesCard = () => {
```
kissmikijr, line 170: Could you pls add tests for this one? It seems this does quiet some things. I think adding some basic ones to check the happy path and sad path would be nice.

plugins/frontend/backstage-plugin-shortcut/src/components/Widgets/StoriesCard.tsx
Irma12: Yeah, tbh I didn't find a way to test them without trying each one of them so I just went straight in with this. I will check 😄

plugins/frontend/backstage-plugin-shortcut/src/api/ShortcutClient.ts
```diff
@@ -0,0 +1,61 @@
...
+import { DiscoveryApi } from '@backstage/core-plugin-api';
+
+const DEFAULT_PROXY_PATH = '/shortcut/api';
+
+export type Options = {
+  discoveryApi: DiscoveryApi;
+  proxyPath?: string;
+};
+
+export class ShortcutClient implements ShortcutApi {
```
kissmikijr, line 28: I think this client would be worth to write some tests for the public functions. It seems there are not much happening but I'd check if it throws an error and fetches from the correct path.

plugins/frontend/backstage-plugin-shortcut/src/components/Widgets/StoriesCard.tsx
kissmikijr: Yea, it is cumbersome, I'd love for our plugins to use the most CSS from the `theme` so when they get implemented in a backstage they would still look consistent with their setup

plugins/frontend/backstage-plugin-shortcut/README.md
kissmikijr: Yea, make sense, let's meet in the middle and add some example to include it in their homepage!

dtuite:
Love this idea

plugins/frontend/backstage-plugin-shortcut/src/mocks/handlers.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import { rest } from 'msw';
+import { storiesResponseMock, usersResponseMock } from './mocks';
+
+export const handlers = [
+  rest.get('http://exampleapi.com/shortcut/api/members', (_, res, ctx) => {
+    return res(ctx.json(usersResponseMock));
+  }),
+  rest.get(
+    'http://exampleapi.com/shortcut/api/search/stories?page_size=25&query=owner:testuser',
```
kissmikijr, line 25: We don't need to provide the query params here in the URL, If we want to handle different mocks for different URL params we can do it something like this:
```
const pageSize = req.url.searchParams.get('page_size');
const query = req.url.searchParams.get('query')
if (query === 'owner:testuser' && pageSize == 25) {
      return res(ctx.json(storiesResponseMock));
}
```

kissmikijr, line 25: We don't need to provide the query params here in the URL, If we want to handle different mocks for different URL params we can do it something like this:
```
const pageSize = req.url.searchParams.get('page_size');
const query = req.url.searchParams.get('query')
if (query === 'owner:testuser' && pageSize == 25) {
      return res(ctx.json(storiesResponseMock));
}
```

kissmikijr (dismissed):

kissmikijr (approved):

PR #498, opened 2022-04-22
Add homepage card for iframe plugin

nicolasarnold12321:
[ch8171]

### DIFF
--- a/plugins/frontend/backstage-plugin-iframe/src/components/HomePageCard.tsx
+++ b/plugins/frontend/backstage-plugin-iframe/src/components/HomePageCard.tsx
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2021 Larder Software Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


plugins/frontend/backstage-plugin-iframe/src/components/HomePageCard.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+      <ErrorComponent {...{errorMessage}}/>
+    )
+  }
+
+  if(errorMessage !== ""){
+    return (
+      <ErrorComponent {...{errorMessage}}/>
+    )
+  }
```
kissmikijr, line 46: Some suplicaton here, what did you want to check in the second if ?

plugins/frontend/backstage-plugin-iframe/src/plugin.ts
```diff
@@ -54,3 +55,18 @@ export const EntityIFrameContent = iframePlugin.provide(
   }),
 );
 
+export const HomePageIFrameCard = homePlugin.provide(
+  createCardExtension<{
+    src: string;
+    title?: string;
```
kissmikijr, line 61: You don't need to give the title prop here or do you use it inside the `./components/HomePageCard` ?

kissmikijr: Ah I see, you use it, so take this as I didn't say anything

plugins/frontend/backstage-plugin-iframe/src/components/HomePageCard.tsx
nicolasarnold12321: silly me. Thank you

plugins/frontend/backstage-plugin-iframe/src/plugin.ts
nicolasarnold12321: I'll remove it!

PR #505, opened 2022-04-27
Rss plugin

punkle:


plugins/home/backstage-plugin-home-rss/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "cross-fetch": "^3.1.4",
+    "msw": "^0.38.2",
+    "react-use": "^17.3.2",
+    "winston": "^3.2.1"
+  },
+  "peerDependencies": {
+    "react": "^16.13.1 || ^17.0.0"
+  },
+  "devDependencies": {
+    "@backstage/cli": "^0.16.0",
```
kissmikijr, line 47: We just bumped the plugins versions to the latest, which will bump the backstage/cli to version 0.17.0 and has potential breaking changes. Could you please bump this package also to the latest ? 
```
yarn backstage-cli versions:bump
```

plugins/home/backstage-plugin-home-rss/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "name": "@roadiehq/backstage-plugin-home-rss",
+  "version": "0.0.1",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "license": "Apache-2.0",
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
```
kissmikijr, line 11: There was introduced a new field to the plugins package.json the `backstage.role` this is a frontend-plugin could you pls add it to this one ?
```
  "backstage": {
    "role": "frontend-plugin"
  },
```

kissmikijr (changes requested):
Just some tiny things to consider

plugins/home/backstage-plugin-home-rss/src/RSSCard/Content.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const [title, setTitle] = useState<string | undefined>();
+
+  useAsync(async () => {
+    if (error) { return }
+    if (data.length > 0) { return }
+    const headers = new Headers({
+      "Accept": "application/rss+xml"
+    });
+    try {
+      const response = await axios.get(props.feedURL, {headers: headers});
```
kissmikijr, line 50: Upstream decided to not use axios, I think we should follow it in our plugins. So in the frontend plugins we could just use the native `fetch`.

kissmikijr (approved):

Issue #502, opened 2022-04-26
Backstage Jira Plugin: Proxy Path in `app-config.yaml` is Ignored


hsparmar1:
The Backstage Jira Plugin has been configure according to the Roadie Documentation found here: https://roadie.io/backstage/plugins/jira/. However, the plugin is failing with the following error in the Backstage UI:

After looking through the modest logging, I found that the `apiUrl` being used is as follows: /api/proxy/**jira/api**/rest/api/latest/project/YYYY. When I cURL using this URL, using the following cURL command

> curl -vvv -X GET -H "Authorization: Bearer XXXXXXXXXXXXXXXXXXXXXXXXXXXX" https://xxx.xxx.com/jira/api/rest/api/latest/project/YYYY

I get the following:

> <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
> <html><head>
> <title>302 Found</title>
> </head><body>
> <h1>Found</h1>
> <p>The document has moved <a href="https://xxx.xxx.com/shibboleth-ds/index.html?entityID=https%3A%2F%2Fxxx.xxx.com&amp;return=https%3A%2F%2Fmaestro.dhs.gov%2FShibboleth.sso%2FLogin%3FSAMLDS%3D1%26target%3Dss%253Amem%253A2269f2bqar5ha3yat4yjy4gzrcXXXXXRGHATAFzfdbaha5ahtw3rf1e6c8d5cb">here</a>.</p>
> </body></html>
> * Connection #0 to host xxx.xxx.com left intact

However, when I use the following cURL command, I get the correct data back:

> curl -vvv -X GET -H "Authorization: Bearer XXXXXXXXXXXXXXXXXXXXXXXXXXXX" https://xxx.xxx.com/**jira**/rest/api/latest/project/YYYY

Notice that the in the second invocation of the cURL command there is no `api` after `jira`. And it delivered the correct results.

The conclusion that I drew here was that the Proxy Path should not be '/jira/api' as it is listed in the Roadie documentation, but should simply be '/jira'. When I change the Proxy Path in the Backstage `app-config.yaml` file to '/jira', 

>   '/jira':
>     target: https://xxx.xxx.com
>     headers:
>       Authorization: Basic XXXXXXXXXXXXXXXXXXXXXXXXX==
>       Accept: 'application/json'
>       Content-Type: 'application/json'
>       X-Atlassian-Token: 'nocheck'
[...]

PR #515, opened 2022-04-29
[SC-8448] - support custom proxy path

kissmikijr:
Fixes the capability to override the proxy path for the plugin

### DIFF
--- a/packages/entities/test-entity.yaml
+++ b/packages/entities/test-entity.yaml
@@ -31,8 +31,8 @@ metadata:
     github.com/project-slug: RoadieHQ/roadie-backstage-plugins
     aws.com/lambda-function-name: HelloWorld
     aws.com/lambda-region: eu-west-1
-    jira/project-key: MIKLOSTEST
-    # jira/component: COMP
+    jira/project-key: TEST
+    jira/component: COMP
     datadoghq.com/dashboard-url: https://p.datadoghq.eu/sb/test-datadog-link
     datadoghq.com/graph-token: qwertytoken1234
     argocd/app-name: test-app
--- a/packages/entities/test-entity.yaml
+++ b/packages/entities/test-entity.yaml
@@ -31,8 +31,8 @@ metadata:
     github.com/project-slug: RoadieHQ/roadie-backstage-plugins
     aws.com/lambda-function-name: HelloWorld
     aws.com/lambda-region: eu-west-1
-    jira/project-key: TEST
-    jira/component: COMP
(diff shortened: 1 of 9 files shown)


packages/entities/test-entity.yaml
```diff
@@ -31,8 +31,8 @@ metadata:
     github.com/project-slug: RoadieHQ/roadie-backstage-plugins
     aws.com/lambda-function-name: HelloWorld
     aws.com/lambda-region: eu-west-1
-    jira/project-key: TEST
-    jira/component: COMP
+    jira/project-key: MIKLOSTEST
+    # jira/component: COMP
```
Irma12, line 35: Is this a typo?

plugins/frontend/backstage-plugin-jira/config.d.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/** Configuration for the Jira plugin */
+export interface Config {
+  jira?: {
+    /**
+     * The proxy path for Jira.
+     * Should be used if proxy is in use for security etc purposes.
+     * @visibility frontend
+     */
+    proxyPath?: string;
```
Irma12, line 25: I am not sure I follow why this proxyPath should be treated different and handled here instead of app-config?

kissmikijr: What do you mean by handled in app-config ? 

The addition of this file to the plugin will make sure we can configure the plugin in the app-config at the `jira` key.

Irma12: I see, what is the value of having custom proxy path instead of predefined (jira/api) one?

Irma12: I found the original issue https://github.com/RoadieHQ/roadie-backstage-plugins/issues/502#issuecomment-1113670104, will check that first

packages/entities/test-entity.yaml
kissmikijr: reverted it back to the original config

plugins/frontend/backstage-plugin-jira/config.d.ts
Irma12: Yes, I think the original issue was due 'target' in 'jira/api' proxy probably being misconfigured or wrong. I don't mind adding this possibility if you want but can we then add an explanation In README?

Irma12 (approved):
Nice

PR #521, opened 2022-05-03
feat: use fetchApi core plugin for Buildkite API client

knksmith57:

### DIFF
--- a/.changeset/moody-beds-bathe.md
+++ b/.changeset/moody-beds-bathe.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-buildkite': patch
+---
+
+Utilize fetchApi core plugin over native fetch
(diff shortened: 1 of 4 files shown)


knksmith57:
@alecjacobs5401 and I are pretty sure this will make the frontend client behave and properly attach authentication to the `fetch()` calls made to `/proxy`. this is probably busted for anyone currently gating backend plugins with auth (like us)

kissmikijr:
Nice spot! Yes it shouldn't work at the current state thanks for this PR.  

Could you pls add a changeset? 
in the root of the repo run:
```bash
yarn changeset
```
I think a patch bump should be good for this one.

And regarding the CI tsc failed because you'll need to add the new dependency in the test also. You should be able to get a   `MockFetchApi` from the `@backstage/test-utils package`

knksmith57:
Thank you for the quick response!

My teammate, @alecjacobs5401, is going to pick this up for us and address your feedback 🙏

alecjacobs5401:
Hey @kissmikijr - made the requested changes you suggested!

kissmikijr (approved):

kissmikijr:
Thank you for this work!

Issue #183, opened 2021-06-24
Filter Activity Stream by Component


crivetechie:
The current behaviour of the plugin is to provide a summary of the opened issues filtering by component (`jira/component`) and project (`jira/project-key`) as well as displaying the activity stream for the whole project.

Would it be possible to make it so that the "Activity stream" shows only the activity for the issues related to the Component identified by the annotation `jira/component` rather than displaying the activity for the whole project `jira/project-key` ?

callmegar:
+1 on this, removing the activity stream is good in this situation, another option could be to add an Epic and pull only issues from a specific Epic into the activity stream

ivangonzalezacuna:
I've been testing an option to filter the activity stream and I think I found a solution for it. 

In our company's JIRA instance we have a filter by `issue-key`, so I changed the issueTypeCounter to a paged request getting only the issue key and the issue type object to reduce the amount of data. Then I use a `.reduce()` to make the count as before and, finally, I store a list of issue-keys. That list will be used to fetch the stream instead of the project key:`streams=issue-key+IS+(list of issue keys)`. I also found out the confluence activities might be filtered out as well, at least for our case.

If you want to, I could create a pull request for that, I guess it's the only possible option to filter the activity stream right now, but only if your company has the `issue-key` filter available.

PR #527, opened 2022-05-05
Feature/filter activity stream by component

ivangonzalezacuna:
As requested in this issue: https://github.com/RoadieHQ/roadie-backstage-plugins/issues/183, when the annotation `jira/component` is defined, the Activity Stream does not filter the activities at all. However, I found a workaround to make that work.

If your JIRA instance has a filter by `issue-key`, then you can requests the changes by ticket or list of tickets.

If the component annotation is not present, then the code is going to be executed exactly in the same way as it was before, but if the component is defined, then the way of counting the issue types and to fetch the activity stream will be slightly different. First of all, the issue type will now return some data (just the `key` and the `issuetype` in order to reduce the amount of processed data) and use a pagination approach. Then, with a `reduce()`, the `IssueCounter` will be generated, and a list of `ticketIds` will be also returned. That list will be then used to fetch the activity stream, instead of the `projectKey`.

Finally, I've added another parameter to the config called `confluenceActivityFilter`, which allows you to filter out from the activity stream the events that happened in confluence and the component is also defined. Apparently the filtering is showing also the entries without `issue-key`, so filtering out all those activities solve that issue. It's described in the README.md how to obtain the key to test it.

I have tested it and it's working as expected. Let me know if you like this solution :smile: 

### DIFF
--- a/.changeset/weak-timers-scream.md
+++ b/.changeset/weak-timers-scream.md
@@ -1,0 +1,7 @@
+---
+'@roadiehq/backstage-plugin-jira': minor
+---
+
+Added a new way to fetch the issues and count them. Added a functionality to filter the activity stream by component name (if present). What's more, if Confluence is also configured and connected to your JIRA instance, you might need to define the `jira.confluenceActivityFilter` variable in your `app-config.yaml` to avoid those activities to be shown in your Activity Stream.
+
+In case the annotation `jira/component` is not present, the plugin will work exactly in the same way as in the previous versions.
--- a/packages/app/cypress/fixtures/jira/searchresult.json
+++ b/packages/app/cypress/fixtures/jira/searchresult.json
@@ -1,6 +1,31 @@
 {
-    "startAt": 0,
-    "maxResults": 0,
-    "total": 1,
-    "issues": []
+  "startAt": 0,
+  "maxResults": 50,
+  "total": 2,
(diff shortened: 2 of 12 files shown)


ivangonzalezacuna:
I'm not sure what failed in the `e2e-tests` pipeline. The tests locally are working fine without errors. I'm on node v16.14.2

kissmikijr:
@ivangonzalezacuna it seems that the cypress test failed for the jira plugin. You can run It the following way locally.
- Start the example app with `yarn dev` in the root folder. 
- cd into the `packages/app` and then run yarn `cy:dev`

I can take a look a little bit later if you'll still need some help with it!

ivangonzalezacuna:
@kissmikijr I have been able to fix one of the issues, but it's still failing. As far as I can tell right now, the intercept of the `activitystream.xml` is failing because the wait returns a timeout error. I tried to open the URL on my browser and I got the expected xml file, so I'm not sure what is going wrong with the test. I haven't worked that much with cypress, so it would be great if you can also give it a check when you get some free time for that.

ivangonzalezacuna:
Hi @kissmikijr , finally solved the problem with the Cypress tests. Now it's ready :smile:

kissmikijr (approved):
Thanks, this looks good!

Issue #516, opened 2022-05-01
Use cross-fetch instead of axios


dweber019:
## Expected Behavior
I would expect that the ArgoCD backend plugin is using cross-fetch to make requests, as it's the default for all plugins within the Backstage ecosystem.

## Current Behavior
Using axios https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/backend/backstage-plugin-argo-cd-backend/package.json#L34

## Possible Solution
Replace axios with cross-fetch

martina-if:
Hi @dweber019 , thank you for raising this. I would like to understand what is the biggest benefit for our users if we change this dependency? Also, is it causing problems in your Backstage instance?

dweber019:
Hi @martina-if 
The users have to set the proxy like `HTTPS_PROXY` but most Backstage use will have https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/help-im-behind-a-corporate-proxy.md. It's an new dependency to manage beside cross-fetch, which does more or less the same.
I think it's the standard to use cross-fetch for Backstage plugins.

PR #539, opened 2022-05-15
feat: use cross-fetch instead of axios for rest requests

dweber019:
Use cross-fetch instead of axios for rest requests.
Will fix https://github.com/RoadieHQ/roadie-backstage-plugins/issues/516

### DIFF
--- a/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
+++ b/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
@@ -1,6 +1,6 @@
-import { Config } from '@backstage/config';
-import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
-import { Logger } from 'winston';
+import {Config} from '@backstage/config';
+import fetch from 'cross-fetch';
+import {Logger} from 'winston';
 
 export type findArgoAppResp = {
   name: string;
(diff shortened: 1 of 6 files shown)


kissmikijr:
It seems there are some mixed tabs and spaces, could you try to run prettier with the project config ?

kissmikijr:
hm maybe not tabs and spaces, but something is off with the formatting. Can you double check that and try to eliminate the formatting changes in this PR?

dweber019:
Reformatted the two files with 2 spaces.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -201,42 +196,39 @@ export class ArgoService implements ArgoServiceApi {
...
-    sourceRepo,
-    destinationServer,
-  }: CreateArgoProjectProps): Promise<object> {
+                            baseUrl,
+                            argoToken,
+                            projectName,
+                            namespace,
+                            sourceRepo,
+                            destinationServer,
+                          }: CreateArgoProjectProps): Promise<object> {
```
kissmikijr, line 231: These are also only formatting changes , right ?

kissmikijr, line 231: These are also only formatting changes , right ?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -257,36 +249,28 @@ export class ArgoService implements ArgoServiceApi {
...
-        throw new Error(
-          'Duplicate project detected. Cannot overwrite existing.',
-        );
-      } else {
-        throw new Error('Could not create ArgoCD project.');
-      }
+    const resp = await fetch(`${baseUrl}/api/v1/projects`, options);
+    const responseData = await resp.json();
+    if (resp.status === 403 || resp.status === 404) {
+      throw new Error(responseData.message);
```
kissmikijr, line 263: We will throw now in cases when it is 404. In the previous version we did return when we got 404. Can you double check it will be ok to throw an Error in case of 404 ?

kissmikijr, line 263: We will throw now in cases when it is 404. In the previous version we did return when we got 404. Can you double check it will be ok to throw an Error in case of 404 ?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -394,9 +372,11 @@ export class ArgoService implements ArgoServiceApi {
   }
 
   async syncArgoApp(
-    {argoInstance,
+    {
+      argoInstance,
       argoToken,
-      appName}: SyncArgoApplicationProps
+      appName
+    }: SyncArgoApplicationProps
```
kissmikijr, line 379: These looks like formatting changes

kissmikijr:
The linter failed on the builds, you'll need to remove an axios import.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
dweber019: Change it back to return the data.

kissmikijr (approved):

PR #548, opened 2022-05-20
Migrate aws-auth plugin

Irma12:
When migration took place we didn't add https://github.com/RoadieHQ/backstage-plugin-aws-auth to monorepo.

### DIFF
--- a/plugins/backend/backstage-plugin-aws-auth/package.json
+++ b/plugins/backend/backstage-plugin-aws-auth/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "@roadiehq/backstage-plugin-aws-auth",
+  "version": "0.3.4",
+  "main": "src/index.ts",
+  "types": "src/index.ts",
+  "license": "Apache-2.0",
+  "private": false,
+  "publishConfig": {
+    "access": "public",
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "scripts": {
+    "start": "backstage-cli backend:dev",
+    "build": "backstage-cli backend:build",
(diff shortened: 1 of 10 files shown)


plugins/backend/backstage-plugin-aws-auth/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@roadiehq/backstage-plugin-aws-auth",
+  "version": "0.3.4",
```
nicolasarnold12321, line 3: Does this need to be bumped?

plugins/backend/backstage-plugin-aws-auth/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "github:RoadieHQ/roadie-backstage-plugins",
+    "directory": "plugins/backend/backstage-plugin-argo-cd-backend"
+  },
+  "dependencies": {
+    "@backstage/backend-common": "^0.9.7",
+    "@types/express": "^4.17.11",
+    "aws-sdk": "^2.902.0",
```
nicolasarnold12321, line 33: We'll probably want to migrate to a later version (eg v3). I'd say it should be another task though

nicolasarnold12321 (approved):

kissmikijr (approved):

plugins/backend/backstage-plugin-aws-auth/package.json
Irma12: Actually we won't be publishing it atm as there no real changes, so it can be bumped when changelog is added, or when real changes does take place :)

PR #555, opened 2022-06-03
feat(useGithubLoggedIn): check for cached token

kuangp:
Optionally check for cached access token in order to set the signed-in state if it exists. This prevents the home page components from showing the `GithubNotAuthorized` component every time Backstage is refreshed if the user is already authorized with GitHub. 

### DIFF
--- a/plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithubLoggedIn.tsx
+++ b/plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithubLoggedIn.tsx
@@ -37,6 +37,7 @@ export const useGithubLoggedIn = () => {
   const [isLoggedIn, setIsLoggedIn] = useState(false);
 
   useEffect(() => {
+    githubApi.getAccessToken('repo', { optional: true });
     const authSubscription = githubApi.sessionState$().subscribe(state => {
       if (state === SessionState.SignedIn) {
         setIsLoggedIn(true);
--- a/.changeset/stupid-donkeys-sell.md
+++ b/.changeset/stupid-donkeys-sell.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-github-pull-requests': patch
+---
+
+Check for cached access token to determine logged in state


plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithubLoggedIn.tsx
```diff
@@ -37,6 +37,7 @@ export const useGithubLoggedIn = () => {
   const [isLoggedIn, setIsLoggedIn] = useState(false);
 
   useEffect(() => {
+    githubApi.getAccessToken('repo', { optional: true });
```
punkle, line 40: Thanks so much for this. :) At the moment I am not fully understanding the change.

Why do we need to retrieve the access token here?
Should we await for the token from the promise?
Should the option also include: `instantPopup: false` so that it does not prompt the user?

As I understand it, we do not want to prompt the user to login but simply tell if the user is loggedin in the component..

kuangp: @punkle here's some more context:

The issue is that `githubApi.sessionState$()` reflects a non signed-in state if nothing has triggered the auth flow yet. This change is meant to optionally trigger the auth flow and check if the user can be considered logged in if a cached access token exists. The result of the request for the access token here does not really matter and we don't need to await it, we just use it to check if this access is already granted which will update the `githubApi.sessionState$()` observable if so. `instantPopup: false` is not necessary since it defaults to `false`. This will not prompt the user. See relevant docs for these options [here](https://github.com/backstage/backstage/blob/d06fbf7bb0d182aab04e70e325e483f5b9f41038/packages/core-plugin-api/src/apis/definitions/auth.ts#L78-L97).

This is the same pattern that Backstage's [`SignInPage`](https://github.com/backstage/backstage/blob/0c31fdb95848211de355ef954529515cd1100342/packages/core-components/src/layout/SignInPage/SignInPage.tsx#L111-L114) performs to check for an existing session so that you're not prompted to re-login everytime the app is refreshed

kissmikijr: Thank you for the explanation! :)

kissmikijr (approved):

plugins/frontend/backstage-plugin-github-pull-requests/src/components/useGithubLoggedIn.tsx
kuangp: @punkle any other concerns here?

punkle: No concerns. Thanks for the explanation.

punkle (approved):

PR #564, opened 2022-06-17
fix(argocd): encode app selectors

### LINKED ISSUE #559: [ArgoCD] Selectors are not properly escaped
The names and selectors of component annotations are not properly escaped which breaks app routing resulting in 404s.

Here's a sample from the `catalog-info.yaml`:

```yaml
apiVersion: backstage.io/v1alpha1
kind: Component
metadata:
  name: my-server
  annotations:
    argocd/app-selector: my.custom.namespace/application-name=my-server
```
 
## Expected Behavior

There should be a request to `/api/argocd/find/selector/my.custom.namespace%2Fapplication-name%3Dmy-server`.

## Current Behavior

There's a request made to `/api/argocd/find/selector/my.custom.namespace/application-name=my-server`. This unescaped slash in the name causes it to miss the backend plugin route: 

https://github.com/RoadieHQ/roadie-backstage-plugins/blob/70f72b47c596764edba3f4e06705dfb0fd37c492/plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts#L81-L84

This results in a 404 since nothing is setup to handle `get` requests to `/find/selector/my.custom.namespace/`

## Possible Solution

* Best practice: Escape the values sent to the backend
* Perhaps the easiest: Change the route to ignore/allow slashes

```diff
-    router.get('/find/selector/:argoAppSelector', async (request, response) => { 
+    router.get('/find/selector/:argoAppSelector*', async (request, response) => { 
```

## Your Environment

- `@roadiehq/backstage-plugin-argo-cd-backend`: `2.4.0`
- `@backstage/integration`: `1.0.0`

## Notes

We have not hit this problem yet, but I would expect the same behavior from an app with a slash in the `name` as well:

https://github.com/RoadieHQ/roadie-backstage-plugins/blob/371b57d7705e99c2e07e8d8b903eb2b251a78cbb/plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts#L42-L45

fearphage:
This fixes #559 by escaping the selectors before they are sent to the backend. There are no changes required for the backend since Express automatically decodes these values.

> NOTE: Express automatically decodes the values in req.params (using decodeURIComponent).

Source: [Express documentation](https://expressjs.com/en/api.html#req.params)

### DIFF
--- a/.changeset/slimy-eyes-raise.md
+++ b/.changeset/slimy-eyes-raise.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-argo-cd': patch
+---
+
+properly escaped app selectors
--- a/plugins/frontend/backstage-plugin-argo-cd/src/api/index.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/api/index.ts
@@ -167,7 +167,7 @@ export class ArgoCDApiClient implements ArgoCDApi {
     const proxyUrl = await this.getBaseUrl();
     const url = options.appName
       ? `${proxyUrl}/find/name/${options.appName}`
-      : `${proxyUrl}/find/selector/${options.appSelector}`;
+      : `${proxyUrl}/find/selector/${encodeURIComponent(options.appSelector)}`;
     return fetch(url, {
       method: 'GET',
       headers: {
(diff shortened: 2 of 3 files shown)


kissmikijr (approved):
Nice thanks!

fearphage:
Any idea when this will be pushed out to NPM in a new version of the plugin? /cc @noisyscanner

PR #591, opened 2022-07-25
use sha256 for the resource names from aws

punkle:
This is a better way to name the resources. They are more likely to
be universally unique across all aws resources across accounts.

This adds in the title, which makes the resources more readable in the
backstage UI also.


Xantier (changes requested):
Looks to be missing few of the providers

Xantier (approved):

PR #592, opened 2022-07-26
adds an entity provider for okta groups and users

punkle:
adds an entity provider for okta groups and users


plugins/backend/catalog-backend-module-okta/src/providers/OktaUserProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+      };
+
+      userResources.push(userEntity);
+    });
+
+    await this.connection.applyMutation({
+      type: 'full',
+      entities: userResources.map(entity => ({
+        entity,
+        locationKey: `okta-user-provider:${this.orgUrl}`,
```
kissmikijr, line 76:
```suggestion
        locationKey: this.getProviderName(),
```

kissmikijr: Most providers use the provider name as their location key

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+      };
+
+      groupResources.push(groupEntity);
+    });
+
+    await this.connection.applyMutation({
+      type: 'full',
+      entities: groupResources.map(entity => ({
+        entity,
+        locationKey: `okta-group-provider:${this.orgUrl}`,
```
kissmikijr, line 81:
```suggestion
        locationKey: this.getProviderName(),
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.ts
```diff
@@ -0,0 +1,85 @@
...
+    const groupResources: GroupEntity[] = [];
+
+    const client = this.getClient();
+
+    const defaultAnnotations = await this.buildDefaultAnnotations();
+
+    await client.listGroups().each(async group => {
+      const members: string[] = [];
+      await group.listUsers().each(user => {
+        members.push(user.id);
```
iain-b, line 54: I'm wondering what the use case is for using the group provider and not the user provider? In the case that we don't use the user provider does it make sense to add members by id here? (these users probably don't exist with that id in the catalog in that case). 

This is very much debatable I'm mostly just curious

punkle: I guess we could add both, no harm. But it also is a bit tidy having the separate. But I could persuaded either way.

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.ts
punkle: done

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.ts
punkle: If we leave it as two providers its hard to know if the group provider is enabled and the user one is now. If the user one is disabled and the group one is enabled it does show the users as members when you look at the yaml, but it doesnt show in the ui.

plugins/backend/catalog-backend-module-okta/README.md
```diff
@@ -0,0 +1,34 @@
...
+  groupProvider.run();
+
+  const { processingEngine, router } = await builder.build();
+  await processingEngine.start();
+
+  // ...
+
+  return router;
+}
+```
```
kissmikijr, line 34: Could you add a couple of sentences how to get the creds to make sure this provider works ?

plugins/backend/catalog-backend-module-okta/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "main": "dist/index.cjs.js",
+    "types": "dist/index.d.ts"
+  },
+  "backstage": {
+    "role": "backend-plugin-module"
+  },
+  "homepage": "https://backstage.io",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/backstage/backstage",
```
kissmikijr, line 20: I think this should be 
```suggestion
    "url": "https://github.com/RoadieHQ/roadie-backstage-plugins",
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.ts
iain-b: Myself and Brian had a chat to say that we can add a facade which wraps both in the future if desired.

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      return {
+        listGroups,
+      };
+    }),
+  };
+});
+
+const logger = createLogger({
+  transports: [new transports.Console({ silent: true })],
+});
```
kissmikijr, line 39: there is a voidlogger in the backend-common package that is being used most of the time in the tests.

```suggestion
import { getVoidLogger } from '@backstage/backend-common';
const logger = getVoidLogger()
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupProvider.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      provider.connect(entityProviderConnection);
+      await provider.run();
+      expect(entityProviderConnection.applyMutation).toBeCalledWith({
+        type: 'full',
+        entities: [],
+      });
+    });
+  });
+
+  describe('where there are is a group', () => {
```
kissmikijr, line 66:
```suggestion
  describe('where there is a group', () => {
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { UserEntity } from '@backstage/catalog-model';
+import * as winston from 'winston';
+import { Config } from '@backstage/config';
+import { OktaEntityProvider } from './OktaEntityProvider';
+
+/**
+ * Provides entities from Okta User service.
+ */
+export class OktaUserProvider extends OktaEntityProvider {
```
kissmikijr, line 25: I think the convention is to call it an entityprovider
```suggestion
export class OktaUserEntityProvider extends OktaEntityProvider {
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserProvider.ts
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright 2021 Larder Software Limited
```
kissmikijr, line 2: I think in the notice the date should be the current one, in this case the year 2022 when this was created.

```suggestion
 * Copyright 2022 Larder Software Limited
```

plugins/backend/catalog-backend-module-okta/README.md
punkle: done.

plugins/backend/catalog-backend-module-okta/package.json
punkle: done.

plugins/backend/catalog-backend-module-okta/config.d.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+export interface Config {
+  integrations?: {
+    /**
+     * Okta configuration
+     */
+    okta?: AccountConfig[];
+  };
+}
```
kissmikijr, line 24: I think the configuration of the provider should be under the catalog.providers key. In this case probably stg like: 
```suggestion
export interface Config {
 catalog?: {
    providers?: {
      /**
     * Okta configuration
     */
      okta?: AccountConfig[];
    }
  };
}
```

kissmikijr (commented):
nice work, just some nits here and there

plugins/backend/catalog-backend-module-okta/config.d.ts
punkle: dunno about the providers part.. what if I want to use the okta credentials somewhere else outside of the catalog.. I think that is why we use the integrations part.. but Im happy either way.

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserProvider.ts
punkle: done

plugins/backend/catalog-backend-module-okta/config.d.ts
kissmikijr: ah, you are correct, I thought at first this accountconfig is provider specific configuration, but now I can see it is responsible to configure the okta creds and such things. So let's leave it as is.

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserProvider.ts
punkle: done

plugins/backend/catalog-backend-module-okta/config.d.ts
punkle: I already changed it to catalog.providers. No harm I think. We can move it if we need later.

plugins/backend/catalog-backend-module-okta/src/providers/OktaEntityProvider.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  public async connect(connection: EntityProviderConnection): Promise<void> {
+    this.connection = connection;
+  }
+
+  protected async buildDefaultAnnotations() {
+    return {
+      [ANNOTATION_LOCATION]: `${this.getProviderName()}:${this.orgUrl}`,
```
kissmikijr, line 56: with the current implementation wouldnt it look strange, that the getProviderName contains the orgUrl already. It would look something like this:

```js
"okta-group:https://my-okta-url.com:https://my-okta-url.com"
```

punkle: done

punkle:
@kissmikijr should have addressed that now. :)

kissmikijr (approved):

PR #620, opened 2022-08-11
support srcFromAnnotation for iframe component

punkle:
support srcFromAnnotation for iframe component

### DIFF
--- a/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
+++ b/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
@@ -16,31 +16,25 @@
 
 import React from 'react';
 import { ErrorComponent } from './ErrorComponent';
-import { IFrameProps } from './types';
+import {
+  IFrameProps,
+  IFrameComponentProps,
+  IFrameFromAnnotationProps,
+} from './types';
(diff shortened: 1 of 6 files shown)


plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
```diff
@@ -16,41 +16,118 @@
...
+
+  if (!src) {
+    return (
+      <ErrorComponent
+        errorMessage={`Failed to get url src from the entity annotation ${srcFromAnnotation}`}
+      />
+    );
+  }
+
+  errorMessage = determineError(src, allowList);
```
iain-b, line 71: I know I'm being a bit annoying about this but we should probably check the src url protocol is not `javascript:` (or that it is http/https) while we're validating (we would get that indirectly if the user sets an allow list but out of the box it's insecure)

sblausten (dismissed):
Nice

plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
punkle: I did something similar, except I am insisting it must be a https url. Can you think of a reason why we might need to support other protocols for the iframe plugin?

iain-b (approved):

PR #630, opened 2022-08-22
allow customization of the scaffolder action names

punkle:
allow customization of the scaffolder action names

### DIFF
--- a/.changeset/hot-pugs-pull.md
+++ b/.changeset/hot-pugs-pull.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
+---
+
+Allow customization of the scaffolder action names.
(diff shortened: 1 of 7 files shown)


sblausten:
Sorry I don't have context on this either - what problem is it solving?

iain-b:
While generally I like the idea of flexibility perhaps we should just have generic names? allowing customization means that templates using these actions won't necessarily be compatible across installations and I think that's something we'd want to consider carefully before proceeding.

punkle:
@iain-b Im kinda worried that we might break someone who is already using these tasks.. maybe I can change their names and release a major version?

punkle:
@sblausten The current names for the scaffolder actions contain `roadiehq:utils` before the action name. There are no other scaffolder tasks that follow this naming pattern. This patch set provides the ability for someone to select the name of the task step to use. It does this in a way that does not break compatability.

PR #641, opened 2022-08-30
Create merge action

iain-b:
Creates an action which merges content into files. Filetype is detected by extension. Json & yaml are supported. 

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/json/merge.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/json/merge.ts
@@ -17,7 +17,9 @@
 import { createTemplateAction } from '@backstage/plugin-scaffolder-backend';
 import { resolveSafeChildPath } from '@backstage/backend-common';
 import fs from 'fs-extra';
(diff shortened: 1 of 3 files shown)


plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/json/merge.ts
```diff
@@ -77,3 +79,73 @@ export function createMergeJSONAction({ actionId }: { actionId?: string }) {
     },
   });
 }
+
+export function createMergeAction() {
+  return createTemplateAction<{ path: string; content: any }>({
+    id: 'roadiehq:utils:merge',
```
punkle, line 85: Im not 100% convinced on the naming of these actions but you are just following the pattern to its good.

punkle (approved):

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/json/merge.ts
iain-b: Yeah neither am I. I considered breaking with the pattern but thought better of it.

Issue #642, opened 2022-08-31
Add more fetch options to cross-fetch request


goququ:
Hi team. I need to add additional option to fetch request. In my case, I want to be able to add prop `credentrials: 'include'`.

## Feature Suggestion

May be we can add some additional parameters to fetch request

## Possible Implementation

```javascript
// plugins/scaffolder-backend-module-http-request/src/actions/run/helpers.ts
export interface HttpOptions {
  url: string;
  method: Methods;
  headers: Headers;
  credentials: 'include' | ... ;
  body?: Body;
}

// plugins/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
const httpOptions: HttpOptions = {
  method: input.method,
  url: queryParams !== '' ? `${url}?${queryParams}` : url,
  headers: input.headers ? (input.headers as Headers) : {},
  body: inputBody,
  credentials: input.credentials
};
```

martina-if:
Closing due to inactivity in #643

PR #643, opened 2022-08-31
[scaffolder-backend-module-http-request] - added credentials prop sup…

goququ:
…port

Hi team, more information about this changes [here](https://github.com/RoadieHQ/roadie-backstage-plugins/issues/642)

### DIFF
--- a/.changeset/dry-mangos-protect.md
+++ b/.changeset/dry-mangos-protect.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': minor
+---
+
+Added ability to pass credentials property to the cross-fetch request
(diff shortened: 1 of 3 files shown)


punkle:
@goququ can I ask why you wouldnt configure the backstage proxy to provide the credentials? This change seems unusual, because these credentials will be presented to the backstage proxy backend which is designed to provide auth to the target service.

goququ:
@punkle thank you for the question. May be I missed something, but when I try to do request to the backstage catalog api I don't use proxy. Isn't it?

punkle:
Yes that is true. You can use this to call any api on the backstage backend. The user's backstage token is provided to the backstage backend by this scaffolder action already. https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts#L141

So in theory the catalog backend should be honouring this token and using it to identify the user who triggered the scaffolder task to start.

For example here on the catalog backend, it is using the authorization token to identify the user https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/service/createRouter.ts#L205

punkle:
@goququ were you able to consider my question?

kissmikijr:
@goququ I'll close this to keep things clean. When you'll have some time again to pick this up feel free to reopen this one!

PR #652, opened 2022-09-05
feat: [sc-10582] upgrade depdendencies as mentioned by rugvip

kissmikijr:
Move the deps to peerDependencies with backward compatibility. Migration guide for this is here:
https://backstage.io/docs/tutorials/react-router-stable-migration#for-plugin-authors


punkle (approved):

punkle:
Do we need changesets for these?

punkle (approved):

PR #678, opened 2022-10-01
feat: consume profile properties from Okta

DerekTBrown:
This PR consumes profile fields from Okta (as opposed to using email for everything).

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
@@ -78,7 +78,7 @@ export class OktaUserEntityProvider extends OktaEntityProvider {
         },
         spec: {
           profile: {
-            displayName: user.profile.email,
+            displayName: user.profile.displayName,
             email: user.profile.email,
           },
           memberOf: [],
--- a/.changeset/witty-snails-rule.md
+++ b/.changeset/witty-snails-rule.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-okta': patch
+---
+
+Changed OktaUserEntityProvider to propogate the Okta displayName field into the Backstage displayName field (instead of email).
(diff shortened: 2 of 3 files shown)


plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
```diff
@@ -78,7 +78,7 @@ export class OktaUserEntityProvider extends OktaEntityProvider {
         },
         spec: {
           profile: {
-            displayName: user.profile.email,
+            displayName: user.profile.displayName,
```
punkle, line 81: Is display guaranteed? Should it default to the email?

Do you think title of the entity should show the display name also?

DerekTBrown: > Is display guaranteed? 

I believe the `displayName` is guaranteed to be present on the `profile` object:
https://github.com/okta/okta-sdk-nodejs/blob/master/src/types/models/UserProfile.d.ts#L28

My read from this [article](https://support.okta.com/help/s/article/User-display-name-does-not-have-a-value?language=en_US) is that the `displayName` defaults to `firstname lastname`, is user configurable, and will default to empty string if not properly configured. 

> Should it default to the email?

I don't have a strong preference, but out of simplicity I think it is fine for `displayName` to match between Okta and Backstage. Users can configure expressions in Okta to set `displayName` to the email by default if they need.

> Do you think title of the entity should show the display name also?

`displayName` is potentially modifiable by Okta users, and therefore can potentially contain garbage (ex. spaces, emojis, things that aren't identifiable). I think `email` is a better cannonicalization of the user at the moment.

punkle (dismissed):

DerekTBrown:
@kissmikijr I added a changeset. I haven't used this tool before- so let me know if I did everything correctly.

kissmikijr:
@DerekTBrown Looks good thank you! Especially kudos for the nice descriptive changeset message! :) I'll merge the PR after the build passes! thank you for your contribution!

kissmikijr (approved):

madianas21:
What happened to this PR? Just wondering. I do not see it in the latest [release](https://github.com/DerekTBrown/roadie-backstage-plugins/blob/1b9ba74e729d24d0d61e86f59a514a4635bd8bf2/plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts#L81)

PR #684, opened 2022-10-07
fix(argo-cd): Encode all app names and selectors

planeiii-te:
This is building on #564 and expanding that practice. Encoding `appNames` probably isn't useful, but it's also not harmful. Figured it'd be better than making another PR later.


kissmikijr (approved):
Thanks for fixing this!

PR #700, opened 2022-11-01
Add support for authenticated Jira proxy via Identity API

robinminto:
The Jira plugin uses the Backstage proxy to forward requests to Jira. As soon as authentication for API requests is enabled (https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/authenticate-api-requests.md), calls to the proxy fail.

This change adds support for the Identity API which allows authenticated requests to the proxy.

### DIFF
--- a/.changeset/poor-papayas-listen.md
+++ b/.changeset/poor-papayas-listen.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-jira': minor
+---
+
+Add support for authenticated Jira proxy via Identity API
--- a/plugins/frontend/backstage-plugin-jira/src/api/index.ts
+++ b/plugins/frontend/backstage-plugin-jira/src/api/index.ts
@@ -18,6 +18,7 @@ import {
   ConfigApi,
   createApiRef,
   DiscoveryApi,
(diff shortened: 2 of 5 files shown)


Xantier (commented):
Thanks @robinminto!

This is a very needed addition to the plugin and we want to merge it in. The way it's currently implemented is technically a breaking change since the identity API needs to initialized in the application using the plugin, whereas previously it wasn't necessary.

I this the reality is that it is very unlikely that there are Backstage instances out there without the IdentityAPI initialized so I think this is probably good to be merged in as is. 

Let me sit on this and think for a day and I'll come back to it. What are your thoughts about the potential disruption?

robinminto:
Sorry, I didn't realise this was a breaking change (and so I only bumped the minor version and didn't add anything to the docs). I'm new to Backstage and my newly created Backstage instance has the IdentityApi by default. Thanks for looking at this.

Xantier (approved):
After thinking about this. I think this will be fine as a minor bump. Backstage introduced the identityApi as a core piece ages ago so I believe we can safely assume that that is present in all apps. 

I will merge and release.

PR #707, opened 2022-11-08
expose the rss and markdown plugins as formal plugins

punkle:
expose the rss and markdown plugins as formal plugins

### DIFF
--- a/.changeset/hungry-humans-applaud.md
+++ b/.changeset/hungry-humans-applaud.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-home-rss': patch
+---
+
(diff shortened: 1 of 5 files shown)


.changeset/hungry-humans-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-home-rss': patch
```
Xantier, line 2: We are post 1.x so these should be minor bump since it's new features.

punkle:
You wont see any change @kissmikijr. The result is identical. This just makes the plugins, formally plugins instead of extending the home page plugin.

I suspect this is causing some issues with out internal dynamic plugin system. But its just a guess at this point.

Issue #723, opened 2022-11-14
Add ability to filter groups and users in okta plugin


punkle:
Add ability to filter groups and users in okta plugin

## Context

Some groups and users should not be loaded from okta. As such it should be possible to filter them somehow.

PR #724, opened 2022-11-14
ability to filter users/groups in okta provider

punkle:
ability to filter users/groups in okta provider

In order to do this, I needed to create a new provider that does both the groups and users together in the same provider. This is because if you filter a group, you will likely want to filter its members. Likewise, if you filter all users from a group you likely want to exclude the group.


packages/backend/src/plugins/catalog.ts
```diff
@@ -90,8 +82,8 @@ export default async function createPlugin(
       fn: async () => {
         await provider.run();
       },
-      frequency: Duration.fromObject({ minutes: 5 }),
-      timeout: Duration.fromObject({ minutes: 10 }),
+      frequency: Duration.fromObject({ minutes: 1 }),
```
kissmikijr, line 85: Can you change it back to at least 5 minutes, 1 minute seems like really frequent and I can imagine it will exhaust the ratelimits quite fast.

punkle:
closes #723

packages/backend/src/plugins/catalog.ts
punkle: good catch, done.

plugins/backend/catalog-backend-module-okta/src/providers/filters/filterOperators.ts
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { FilterOperator } from '../../types';
+
+export const filterOperators: Record<
+  FilterOperator,
+  (lhs: any, rhs: any) => boolean
+> = {
+  equals: (lhs: any, rhs: any) => {
```
kissmikijr, line 22: What is lhs, rhs ?

punkle: left had side and right hand side. When referring to operators, you can call each side of the operator by the side its on.

e.g.

`1 < 2`

In this case, the left hand side is 1, and right hand side is 2

I am happy to give them different names if you have better names?

plugins/backend/catalog-backend-module-okta/config.d.ts
```diff
@@ -14,13 +14,17 @@
...
 export interface Config {
   catalog?: {
     providers?: {
       /**
        * Okta configuration
        */
-      okta?: AccountConfig[];
+      okta?: {
+        [authEnv: string]: AccountConfig[];
+      };
```
kissmikijr, line 27: This change is a breaking change, isn't it? I wonder how should we handle this one to hurt the existing users as little as possible

plugins/backend/catalog-backend-module-okta/README.md
```diff
@@ -5,17 +5,104 @@ entity providers to read Okta Group and User objects as Backstage Entities.
...
+import { OktaOrgEntityProvider } from '@roadiehq/catalog-backend-module-okta';
+
+export default async function createPlugin(
+  env: PluginEnvironment,
+): Promise<Router> {
+  const builder = await CatalogBuilder.create(env);
+
+  const oktaConfig = env.config.getOptionalConfigArray(
+    'catalog.providers.okta',
+  );
```
kissmikijr, line 73: I think that this piece of logic should be the responsibility of the provider's static fromConfig method, otherwise it assumes it gets a Config object which has okta specific keys on it.

plugins/backend/catalog-backend-module-okta/src/providers/filters/includeUser.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { User } from '@okta/okta-sdk-nodejs';
+import { UserFilter } from '../../types';
+import { get } from 'lodash';
+import { filterOperators } from './filterOperators';
+
+export const includeUser = (
```
kissmikijr, line 21: Could you add a test case for this function?

plugins/backend/catalog-backend-module-okta/src/providers/filters/includeGroup.ts
```diff
@@ -0,0 +1,31 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { Group } from '@okta/okta-sdk-nodejs';
+import { GroupFilter } from '../../types';
+import { get } from 'lodash';
+import { filterOperators } from './filterOperators';
+
+export const includeGroup = (
```
kissmikijr, line 21: Could you add a test case for this function?

plugins/backend/catalog-backend-module-okta/config.d.ts
punkle: Actually this change is incorrect, Ill fix it.

punkle: Done

punkle: Done

plugins/backend/catalog-backend-module-okta/README.md
punkle: Done

plugins/backend/catalog-backend-module-okta/src/providers/filters/includeUser.ts
punkle: Done

plugins/backend/catalog-backend-module-okta/src/providers/filters/includeGroup.ts
punkle: Done

plugins/backend/catalog-backend-module-okta/src/providers/filters/filterOperators.ts
punkle: @kissmikijr does this answer your question?

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -55,17 +51,9 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
     const orgUrl = config.getString('orgUrl');
     const token = config.getString('token');
 
-    const userFilters = userFiltersFromConfigArray(
-      config.getOptionalConfigArray('userFilters'),
-    );
-    const groupFilters = userFiltersFromConfigArray(
-      config.getOptionalConfigArray('groupFilters'),
-    );
+    const groupFilter = config.getOptionalString('groupFilter');
```
kissmikijr, line 54: the config here will be the root config now, no ? So I think these needs to be something like: `catalog.providers.okta.groupFilter`

punkle: I only changed the root config for the new one I have created. I didn't change this. It would be a breaking change to change this for the existing ones.

I would prefer not to change this because its not within the scope of what I am doing here. I can revisit it after as part of a separate effort if this is what we want?

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+        const groupFilter = oktaConfig.getOptionalString('groupFilter');
+
+        return { orgUrl, token, groupFilter, userFilter };
+      });
+
+    return new OktaOrgEntityProvider(oktaConfigs || [], options);
+  }
+
+  constructor(
+    accountConfig: AccountConfig[],
```
kissmikijr, line 66:
```suggestion
    accountConfigs: AccountConfig[],
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+    options: {
+      logger: winston.Logger;
+      groupNamingStrategy?: GroupNamingStrategies;
+      userNamingStrategy?: UserNamingStrategies;
+    },
+  ) {
+    super(accountConfig, options);
```
kissmikijr, line 73:
```suggestion
    super(accountConfigs, options);
```

kissmikijr (dismissed):

kissmikijr (approved):

PR #725, opened 2022-11-14
Bugfix/namespaced selectors

kschjeld:
Finding an application across multiple backends should not spam the logs with errors, as you would normally expect the app not to be present in all the configured clusters. Currently, the code is not handling return objects from Argo CD when using a selector and nothings is found, causing error when trying to iterate over an undefined map.

When using a selector to find instances, the selector must be compatible with normal label names in Kubernetes, typically on the form prefix/label name. This is not supported as is, as Express will not be able to match the request to the correct GET-method when the argoAppSelector parameter contains slash. Even if the front-end plugin is fixed to handle this, it would make a cleaner API if the selector can be defined directly without encoding it first. In line with the corresponding API provided by ArgoCD.


plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -62,11 +62,10 @@ export class ArgoService implements ArgoServiceApi {
...
         } catch (error: any) {
-          this.logger.error(
-            `failed to fetch app data from ${argoInstance.name}: ${String(
-              error,
-            )}`,
-          );
+          return null;
+        }
+
+        if (options.selector && !getArgoAppDataResp.items) {
```
punkle, line 68: I wonder does it make senses to add a test for this?

kschjeld: There should be tests around this, yes. I will have a look at it and update.

fearphage:
Wasn't this addressed in #564?

Looking at the tests, `/find/selector/prefix/name=test-app` would reach the backend as `/find/selector/prefix%2Fname%3Dtest-app`, no?

kschjeld:
ArgoCD will return the following when not finding anything using a selector:
`{"metadata":{"resourceVersion":"251565525"},"items":null}`

Which is why my fix does an extra check to see if `getArgoAppDataResp.items` is defined if using a selector, and skipping the block trying to map over the items if not.

kschjeld:
So, I have might have put too many fixes into this PR. 

Pri 1 is fixing the code related to item and mapping. 
Pri 2 is removing error-logging for cases that are not really errors - a find returning zero results is normally not an error. 
Pri 3 is removing the need for URL-encoding selectors following the standard Kubernetes standard. I find it to be a strange requirement, compared to the ArgoCD API for example.

But - I can split it into several parts or remove some of it. We are really dependent on getting the pri 1 out and into our environment, and would also much appreciate avoiding spamming the logs with error-messages.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.test.ts
```diff
@@ -73,6 +77,125 @@ describe('router', () => {
...
+      getArgoAppData.mockRejectedValue('error getting application');
+
+      const response = await request(app).get('/find/name/test-app');
+
+      expect(getArgoToken).toBeCalledTimes(2);
+      expect(response.body).toMatchObject([]);
+    });
+
+    it(`successfully finds application based on selector`, async () => {
+      getArgoAppData.mockResolvedValue({
```
fearphage, line 132: I would confirm that `getArgoAppData` is called with the input that you expect.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
```diff
@@ -80,10 +80,13 @@ export function createRouter({
     },
   );
 
-  router.get('/find/selector/:argoAppSelector', async (request, response) => {
-    const argoAppSelector = request.params.argoAppSelector;
-    response.send(await argoSvc.findArgoApp({ selector: argoAppSelector }));
-  });
+  router.get(
+    '/find/selector/:argoAppSelector([^/]*)',
```
fearphage, line 84: This regex looks off. In the case of the test data `/find/selector/prefix/name=test-app`, doesn't this call `findArgoApp` with `{ selector: "prefix" }`? I think you want to use `.+` for the regex. You should add a test to ensure this question is answered.

fearphage:
>Pri 3 is removing the need for URL-encoding selectors following the standard Kubernetes standard. I find it to be a strange requirement, compared to the ArgoCD API for example.

Are you using the backend plugin without the front-end plugin? The front-end plugin handles all this passively and requires no data changes from you.

fearphage (changes requested):

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
kschjeld: I have added extra verifications in the tests to ensure that the URL is parsed correctly when using arbitrary selectors, and that the underlying method is called with the correct input.

planeiii-te (approved):

fearphage (approved):

fearphage:
@kschjeld Are you using the backend plugin without the front-end plugin? The front-end plugin handles all this passively and requires no data changes from you.

kschjeld:
> 

Yes, @fearphage , working on a small plugin to just show some selected info from Argo as an overview.

kschjeld:
The reg exp has bothered me. Seems like different versions of Express handles this differently as well. So have prepared an update to revert back to original router-code and fixed the unit test to URI-encode according to how the frond-end plugin works. As said before, the URI-stuff is not important, we just need this to work with multiple ArgoCD-servers.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.test.ts
```diff
@@ -73,6 +77,153 @@ describe('router', () => {
...
+      const selector = encodeURIComponent(
+        'prefix/name=test-app,other-prefix/name=some-value',
+      );
+      const response = await request(app).get(`/find/selector/${selector}`);
+
+      expect(getArgoAppData).toBeCalledTimes(2);
+      expect(getArgoAppData).toBeCalledWith(
+        'https://argoInstance1.com',
+        'argoInstance1',
+        { selector: 'prefix/name=test-app,other-prefix/name=some-value' },
```
fearphage, line 164: MINOR: A suggestion to make it obvious that the `selector` value is identical.

```suggestion
      const selector = 'prefix/name=test-app,other-prefix/name=some-value';
  
      const response = await request(app).get(`/find/selector/${
        encodeURIComponent(selector as string)
      }`);

      expect(getArgoAppData).toBeCalledTimes(2);
      expect(getArgoAppData).toBeCalledWith(
        'https://argoInstance1.com',
        'argoInstance1',
        { selector },
```

fearphage (dismissed):
A minor suggestion

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.test.ts
kschjeld: Thx. You are of course right. Fixed and pushed.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
kissmikijr, line 68: I'm not familiar with the API. The second part check if the items key exists on the getArgoAppDataResp ? Is this the intended check or you'd like to check if the items array is actually empty? or both ?

kissmikijr (commented):
1 small question

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
kschjeld: Yes. That is intended. When using selectors, the items-property of the returned JSON-object will be undefined, not an empty array. So in that case, trying to iterate over the elements of items would yield nothing but an error.

kissmikijr (approved):
Thank you for handling this edge case, and kudos for adding a test for it! :) Let's ship it!

PR #748, opened 2022-12-01
add yq step action for scaffolder

punkle:
add yq step action for scaffolder

### DIFF
--- a/.changeset/neat-sheep-rest.md
+++ b/.changeset/neat-sheep-rest.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
+---
+
+Add `yq` action to the scaffolder `utils`.
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2021 Larder Software Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createTemplateAction } from '@backstage/plugin-scaffolder-backend';
+import { resolveSafeChildPath } from '@backstage/backend-common';
+import { execFile, ChildProcess } from 'child_process';
+import { promisify } from 'util';
+import fs from 'fs';
+
+const execFilePromise = promisify(execFile);
+
+const createMessageSanitizer = (ctx: { workspacePath: string }) => {
+  return (message: string) => {
+    return message.replace(ctx.workspacePath, '<redacted>');
+  };
+};
+
+const wait = async (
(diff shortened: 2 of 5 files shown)


plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
```diff
@@ -0,0 +1,130 @@
...
+      ctx.logger.info(`Attempting to yq with ${ctx.input.path}`);
+
+      const writePath =
+        ctx.input.writeOutputPath &&
+        resolveSafeChildPath(ctx.workspacePath, ctx.input.writeOutputPath);
+      if (!fs.existsSync(sourceFilepath)) {
+        throw new Error(`${messageSanitizer(sourceFilepath)} does not exist`);
+      }
+      try {
+        const initialResult = await execFilePromise('yq', [sourceFilepath]);
```
punkle, line 109: This type of code that depends on a cli command is very hard to test.

.changeset/neat-sheep-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
```
Xantier, line 2: Is this a breaking change? It introduces an underlying command `yq` which is not one of the default ones in any sane distros

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
Xantier: Does it have to be a cli command?

kissmikijr (changes requested):
Can you elaborate a bit why do we need this command ? In general I don't like that we rely on a cli that needs to be installed on the host where this action runs. 
I think we should be able to do this with native js without other dependencies to be present on the system

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
punkle: I havent found a library for this yet. Suggestions welcome.

.changeset/neat-sheep-rest.md
punkle: My thinking here is that you dont have to choose to use this action and current usage is not using it. So it is not breaking in that sense. Happy to make it breaking if you have a different perspective.

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
Xantier: `yaml` on NPM would allow you to achieve this

punkle:
The use case: I want to add an item to an array in a yaml file and then write it back to the file. I tried lodash.mergeWith but it does not work with nested items. This was suggested that it would be a richer solution in that it would allow lots of other types of interaction with yaml data.

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/yq/yq.ts
punkle: I think its the merging nested arrays part that Ive trouble with. How would that be surfaced via an action?

e.g.

if I have a yaml file containing:

```yaml
person:
   numbers:
    - 555123456
    - 555123457
```

How do I form an action to append a new number to that list and write it back to the file?

With the action I have just created I can do that with:

```yaml
    - id: yamlmerge
      name: YAML Merge
      action: roadiehq:utils:yq
      input:
        path: person.yaml
        writeOutputPath: person.yaml
        query:
          - '(.person.numbers = .person.numbers + 555123458)'
```

punkle: It also allows for deleting that phone number later.

PR #759, opened 2022-12-07
[Feat] scaffolder http-requests module: Add baseUrl override option

sfc-gh-jtsuyemura:
## Problem
The current _**scaffolder-backend-module-http-request**_ module constructs the backend url using the `generateBackstageUrl()` function. This function uses the `backend.baseUrl` defined in the app-config.yaml.

There are some deployment models (E.g. k8s, or separated frontend and backend containers) where different services, routing, dns, etc., requires `backend.baseUrl` in the `app-config.yaml` be different then the URL _**scaffolder-backend-module-http-request**_ module requires.

## Proposed Change

This change allows the user to specify an optional `baseUrl` parameter in the template action to override the default `backend.baseUrl` in `app-config.yaml`.

Example action using `baseUrl`
```
....
  steps:
    - id: some_request_id
      name: Custom HTTP Request - with Base Url Override
      action: http:customHttp:request
      input:
        method: 'GET'
        path: '/api/my-api/health'
        baseUrl: 'http://localhost:3000'    <------ New optional Input
```

- The constructed URL in this case will be: `http://localhost:3000/api/my-api/health`

- If no `baseUrl` is provided, the constructed URL will use the _**backend.baseUrl**_ from _**app-config.yaml**_:  `http://${backend.baseUrl}/api/my-api/health`

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/.changeset/twenty-parents-hope.md
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/.changeset/twenty-parents-hope.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': minor
+---
+
+Added the optional baseUrl scaffolder action property. Used to optionally override backend.baseUrl set in the application config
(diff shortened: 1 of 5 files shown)


punkle:
We have intentionally not allowed this because it allows the template access to any resource accessible to the backend. e.g. a template could in theory be constructed that could access network resources within the reach of the backend service. e.g. database services internal kubernetes apis, aws metadata service etc.

The way it is currently configured should allow you to create a proxy configuration in backstage to connect to your service at whatever baseUrl you might want. If the proxy does not provide enough flexibility, you could create a custom backend plugin to proxy the traffic in whatever way you would prefer.

We do appreciate your contribution here, but unfortunately we will not be able to accept this change at this time.

PR #768, opened 2022-12-12
Use to Environment Variable Substitution in README

madianas21:
As per [this](https://backstage.io/docs/conf/writing#env-includes) official Backstage documentation, [Environment Variable Substitution](https://backstage.io/docs/conf/writing#environment-variable-substitution) is a more convenient way to supply variables. Also, it is much more beginner friendly. The `$env` include is mentioned much later in the documentation and it's not used in any of the examples on PostgreSQL and GitHub configuration, so it looks like it's not a standard way of supplying environment variables.
Using `$env` is not wrong, however, it would save a lot of headache, for anyone only starting with Backstage.

### DIFF
--- a/.changeset/hungry-numbers-hammer.md
+++ b/.changeset/hungry-numbers-hammer.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-okta': patch
+---
+
+Use environment variable substitution in docs example.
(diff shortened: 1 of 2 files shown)


kissmikijr:
Thanks, I agree! Can you please add a patch changeset, just to reflect this readme change in npm also for the package.

kissmikijr (approved):

PR #774, opened 2022-12-14
aggressively remove html comments from markdown content

punkle:
Previously we had been allowing configuration to opt into removing html comments from the markdown. On consideration, I can think of no reason for these to be left in place, so I am intentionally purging them from the markdown content before rendering.


iain-b (changes requested):
While I think it should definitely be the default to remove these what if the document includes a html example (say in a code block) which includes comments? It's an edge case but I think we should keep the option to not purge (but change the default).

punkle:
@iain-b done. thanks.

punkle:
Superceded by #777

PR #776, opened 2022-12-15
Log Okta user/group/org provider completion

dhoizner:
This pull emits an info-level log when the various Okta entity providers complete execution in addition to the existing log that gets emitted when they begin execution. This allows better visibility into provider runtime.

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
@@ -112,5 +112,6 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
         locationKey: this.getProviderName(),
       })),
     });
+    this.logger.info(`Finished providing okta group resources from okta: ${this.orgUrl}`);
   }
 }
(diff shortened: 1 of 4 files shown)


plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -112,5 +112,6 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
         locationKey: this.getProviderName(),
       })),
     });
+    this.logger.info(`Finished providing okta group resources from okta: ${this.orgUrl}`);
```
punkle, line 115: would it make sense to show the number of groups provided?

dhoizner: yup, i think so! updated all three completion logs to include counts of provided users/groups

punkle (approved):

punkle:
@dhoizner thank you. That is released now.

dhoizner:
> @dhoizner thank you. That is released now.

Sweet, thank you!

PR #777, opened 2022-12-15
allow private images be shown on the markdown card

punkle:
allow private images be shown on the markdown card


plugins/home/backstage-plugin-home-markdown/src/apis/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    for (const href of mediaLinks) {
+      const mimeType = mimeTypeLookup(href);
+
+      if (mimeType) {
+        const contentResponse = await octokit.request(
+          `GET /repos/${owner}/${repo}/contents/${href}`,
+        );
+        media[
+          href
+        ] = `data:${mimeType};base64,${contentResponse.data.content.replaceAll(
```
punkle, line 110: This is the magic. It's creating src urls with the data protocol. I.e. inline image src.

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
```diff
@@ -14,34 +14,80 @@
...
   }
 
-  let content = Buffer.from(value.content, 'base64').toString('utf8');
+  if (!value) {
+    return <Progress />;
+  }
 
-  if (props.purgeHtmlComments) {
+  let content = value.content;
+  if (!dontStripHtmlCommentsBeforeRendering) {
```
Xantier, line 80: maybe something like `preserveHtmlComments` or just defaulting the original to false might be better here to avoid the double negation. Also easier to default things to false instead of true

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/useGithubFile.ts
```diff
@@ -29,7 +29,7 @@ export const useGithubFile = (options: MarkdownContentProps) => {
       const octokit = new Octokit({ auth: token });
 
       const response = await octokit.request(
-        `GET /repos/{owner}/{repo}/contents/${path}`,
+        `GET /repos/${owner}/${repo}/contents/${path}`,
```
Xantier, line 32: This mixes JS template string and the templating functionality of the octokit library. Octokit uses `{something}` and expects an object with `{something: 'the-val-of-something'}`, no dollar sign like JS template strings. We should use one or the other but not both

Xantier: Similarly in other spots as well.

punkle: ahhh I was wondering how that even worked.

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
```diff
@@ -14,34 +14,80 @@
...
 
 const GithubFileContent = (props: MarkdownContentProps) => {
-  const { value, loading, error } = useGithubFile({ ...props });
+  const { dontStripHtmlCommentsBeforeRendering } = props;
+  const apiHolder = useApiHolder();
+
+  const { value, loading, error } = useAsync(async () => {
+    const githubClient = getGithubClient(apiHolder);
+    return githubClient.getContent({ ...props });
+  });
```
Xantier, line 53: We want to add the dependency array here also.

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/types.ts
```diff
@@ -24,7 +24,7 @@ export type MarkdownContentProps = {
   repo: string;
   path: string;
   branch?: string;
-  purgeHtmlComments?: boolean;
+  dontStripHtmlCommentsBeforeRendering?: boolean;
```
martina-if: can this break customer's tenants?

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
```diff
@@ -14,34 +14,80 @@
...
+const getGithubClient = (apiHolder: ApiHolder) => {
+  let githubClient: GithubApi | undefined = apiHolder.get(githubApiRef);
+  if (!githubClient) {
+    const auth = apiHolder.get(githubAuthApiRef);
+    if (auth) {
+      githubClient = new GithubClient({ githubAuthApi: auth });
+    }
+  }
+  if (!githubClient) {
+    throw new Error(`Failed to get the github client`);
```
martina-if, line 41: Probably good to add the name of the plugin to the error for traceability when we read these?

martina-if (commented):
I guess we are only transforming urls, right? Are there any security concerns in terms of code injection we should be aware of in your opinion?

Xantier (dismissed):

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/types.ts
punkle: No, not really. its very specific and most people will want these comments gone.

plugins/home/backstage-plugin-home-markdown/src/MarkdownCard/Content.tsx
punkle: good point.

punkle:
@martina-if regarding code injection. I dont know of any code injection concerns. I think its safe though. We are base64 encoding the content of the images and setting them in the src tag of the img component. So I think the possibilities are probably limited. If the image is absolutely enourmous it might slow down the plugin or maybe the app itself. But I think github limits the size of files.

Irma12 (approved):

PR #781, opened 2022-12-19
[iframe plugin] passing `class` props to components

planeiii-te:
The types already permit the `class` prop on the iframe components, but it was not being used. This passes the prop down to the appropriate components to enable styling the iframe. 

I also moved one of the types from inline into the `types.ts`.

### DIFF
--- a/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
+++ b/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
@@ -112,7 +112,13 @@ export const IFrameCard = (props: IFrameComponentProps) => {
   const { src, srcFromAnnotation, height, width, title } = props;
   if (src) {
     return (
-      <IFrameFromSrc class={props.class} src={src} height={height} width={width} title={title} />
+      <IFrameFromSrc
+        class={props.class}
+        src={src}
+        height={height}
+        width={width}
+        title={title}
+      />
     );
   }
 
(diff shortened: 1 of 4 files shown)


plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
```diff
@@ -26,20 +27,14 @@ import { configApiRef, useApi } from '@backstage/core-plugin-api';
...
-type IFrameComponentContentProps = {
-  src: string;
-  title: string;
-  height?: string;
-  width?: string;
-};
-
 const IFrameComponentContent = (props: IFrameComponentContentProps) => {
-  const { title, src, height, width } = props;
+  const { class: className, height, src, title, width } = props;
```
planeiii-te, line 31: Note: `class` is a valid property name, but an invalid variable name due to it being a reserved keyword.

Xantier: can we name this `classes` or `extraClasses` instead of using reserved keywords (`class` in JS, `className` in React)

planeiii-te: `className` is not a reserved word and it makes it explicit to where this will be used was my thinking. I can rename this to whatever though.

Xantier: Sorry I meant the incoming prop in here. We shouldn't be using the `class` keyword as a prop name since it's confusing with JSX JS class and HTML class attribute naming convention mixups.

planeiii-te: I switched it to using `classes` instead.

Xantier (approved):
Nice! Thanks for the contribution. Merging.

PR #786, opened 2022-12-20
refactor markdown plugin to reuse github insights

punkle:
refactor markdown plugin to reuse github insights


Xantier:
IMO it would be better if those shared clients are separate libraries within this repository. The homepage markdown plugin does not convey hard dependency to GitHub or any other SCM and now the whole plugin has changed into HomePage-GitHub-Markdown. It would be much better to be able to change the implementation that provides the actual markdown content (not rendering)

punkle:
The markdown plugin only ever supported markdown hosted on github. So in a way it was always the github-markdown plugin. I agree that it should support multiple providers. In fact it probably should. I had duplcated all of the api and component code from the markdown plugin (for the fix for private images) into the readme component. I am not sure what the best way of removing this duplication is.
Do you know is there a fromtend api that does the same as the url readers in the backend?

punkle:
We have decided that the duplication of code is acceptable for the moment. We can make the markdown plugin support mulitple markdown sources at a later point.

PR #815, opened 2023-01-11
Support additional argocd label constraints

wpessers:
Removed hardcoded '=' constraint from url query parameters. Added support for different argocd label constraints


punkle:
Why does something like this not work:

```
metadata:
  annotations:
    argocd/app-selector: "environment notin (dev,staging)"
```

wpessers:
> Why does something like this not work:
> 
> ```
> metadata:
>   annotations:
>     argocd/app-selector: "environment notin (dev,staging)"
> ```

The argocd api does not really follow the same syntax as the argocd cli. But I can alter the code to support the annotation in the way you wrote it here if you want.

PR #821, opened 2023-01-19
add 'user-agent' header to cloudsmith api call and update readme

BartoszBlizniak:
- added "User-Agent" header to the Cloudsmith API request
- updated README to include the header
- added Cloudsmith to the plugin list in the main README

### DIFF
--- a/plugins/frontend/backstage-plugin-cloudsmith/README.md
+++ b/plugins/frontend/backstage-plugin-cloudsmith/README.md
@@ -57,6 +57,7 @@ proxy:
     target: 'https://api.cloudsmith.io/v1'
     headers:
       X-Api-Key: ${CLOUDSMITH_API_KEY}
+      User-Agent: 'Roadie-Backstage'
 ```
 
 When you run the backstage backend, you will need to set the `CLOUDSMITH_API_KEY` environment variable.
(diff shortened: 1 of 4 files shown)


plugins/frontend/backstage-plugin-cloudsmith/README.md
```diff
@@ -57,6 +57,7 @@ proxy:
     target: 'https://api.cloudsmith.io/v1'
     headers:
       X-Api-Key: ${CLOUDSMITH_API_KEY}
+      User-Agent: 'Roadie-Backstage'
```
punkle, line 60: This should be just Backstage. It can be used by backstage deployments other than roadie.

BartoszBlizniak:
Thanks @punkle , I have also sent a DM on Discord if you have time to check at some point, thanks!

kissmikijr (approved):

punkle:
@BartoszBlizniak just a note. I think it woul dbe preferable for your self if you set that header at runtime in the CloudsmithClient. That way you dont have to rely on the special configuration of the proxy for it to be set.

punkle:
Also if you want to create a CloudsmithClient for javascript package and use this in this plugin with the user agent set to whatever you like its going to be even better.

PR #827, opened 2023-01-26
feat: Add option to filter alerts by labels

mimotej:
This PR introduces a new annotation called `prometheus.io/labels`. This annotation is optional, and when used, it filters alerts by labels that are present. It filters an array of alerts created using a `prometheus.io/alert` annotation. Labels are defined in the following format:
```
prometheus.io/labels: "label1=value1,label2=value2"
```
Alerts have to have all the labels defined, or they are not included in the final array.


.changeset/light-melons-share.md
```diff
@@ -0,0 +1,11 @@
+---
+'@roadiehq/backstage-plugin-prometheus': major
```
Xantier, line 2: This doesn't look like a breaking change so a `minor` bump is enough when introducing new functionality I think

Xantier (dismissed):
Excellent. Thanks for the contribution. I'll merge this in after CI and after the change to a minor release

.changeset/light-melons-share.md
mimotej: Good point :+1:. Fixed thank you

Xantier (approved):
:shipit:

PR #841, opened 2023-02-09
feat: allow setting the baseUrl in the scaffolder-backend-module-http-request

alexef:
The backstage http plugin is great, but running backstage behind the oauth2Proxy makes authentication from the task cumbersome. 
We would like to be able to configure the baseUrl to http://localhost:7007 as task runner and backend are collocated.

Signed-off-by: Alex Eftimie <alex.eftimie@getyourguide.com>

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.ts
@@ -23,7 +23,9 @@ const DEFAULT_TIMEOUT = 60_000;
 
 export const generateBackstageUrl = (config: Config, path: string): string => {
   // ensure the request points to the correct domain
-  const externalUrl = config.getOptionalString('backend.baseUrl') || '';
+  const externalUrl = (
+    config.getOptionalString('plugin.scaffolder-backend-module-http-request.baseUrl') || config.getOptionalString('backend.baseUrl') || ''
+  );
   if (externalUrl === '') {
     throw new Error('Unable to get base url');
   }
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/README.md
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/README.md
@@ -138,3 +138,8 @@ Headers
 ```
 
 You can also visit the `/create/actions` route in your Backstage application to find out more about the parameters this action accepts when it's installed to configure how you like.
+
+## Configuration options
+
+By default, the plugin will use `backend.baseUrl` as the Backstage backend address. 
+This can be overriden by setting `plugin.scaffolder-backend-module-http-request.baseUrl`.
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.test.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.test.ts
@@ -25,6 +25,7 @@ import { Writable } from 'stream';
 import * as winston from 'winston';
 
 const mockBaseUrl = 'http://backstage.tests';
+const mockCustomBaseUrl = 'http://localhost:7007';
 
 let mockResponse: Response;
 let config: Config;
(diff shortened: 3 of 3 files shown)


plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.ts
```diff
@@ -23,7 +23,9 @@ const DEFAULT_TIMEOUT = 60_000;
 
 export const generateBackstageUrl = (config: Config, path: string): string => {
   // ensure the request points to the correct domain
-  const externalUrl = config.getOptionalString('backend.baseUrl') || '';
+  const externalUrl = (
```
punkle, line 26: I wonder if we instead passed in an implementation of the DiscoveryApi, then you could implement it how you like, and simplify this function at the same time.

This function would end up being something like this:

```
export const generateBackstageUrl = (path: string) {
  discovery.getBaseUrl(path)
}
```

This would require an api change to `createHttpBackstageAction`, but maybe it can be done in a compatible way.. thoughts?

alexef: I can give it a try, yes. 
to stay compatible, we could pass both an optional discovery and config to `createHttpBackstageAction`, but is that really needed? the only use of `config` is here, so I would prefer to break backwards compatibility and pass only the discovery

punkle: ok. cool. Sounds ok to me. We will just need to make sure that the major version is bumped to indicate that there is a breaking api change. What do you yhink @kissmikijr

alexef: @punkle have a look: https://github.com/RoadieHQ/roadie-backstage-plugins/pull/842 - I'm not really happy with the outcome.

Of course we can provide our own implmenetation of discoveryApi instead of `UrlPatternDiscovery`, but it still feels like we're forcing an API (url + path) on something that was meant for plugins: (url + pluginName).

punkle: I might not be understanding the problem you are trying to solve. The backstage backend should be discoverable via the discovery api. Could you elaborate on what you are trying to do?

alexef: this plugin supports http requests to the Backstage backend, by only specifying the path (and building the url using `backend.baseUrl`), for instance:
`path: /api/proxy/my-backend` results in `https://backstage.example.com/api/proxy/my-backend`

because my backend requests are intercepted by Oauth2Proxy, I would like the plugin to make the requests against a different url than `backend.baseUrl`, in the same example: `path: /api/proxy/my-backend` results in `http://localhost:7007/api/proxy/my-backend`.

note that _the paths have nothing to do with the http action plugin, they are paths to various endpoints and various plugins in the backstage backend_.

the DiscoveryApi and the DiscoveryService are meant to build url bases for a plugin. in my example, if the plugin name is `myplugin`, then such an api would normally build: `https://backstage.example.com/api/plugin/myplugin/`, so I can use this baseUrl and build other urls _to the plugin itself_.  this is where I see the mismatch. on one hand, I'm building paths to anything on the other I'm using a discovery api to figure out the base path for the plugin itself. 

hope the examples make sense

alexef: I just realized I can do this instead in `scaffolder.ts`:

``` 
const httpConfig = new ConfigReader({
    backend: {
      baseUrl: "http://localhost:7007"
    }
  });
  const actions = [
    createHttpBackstageAction({ config: httpConfig }),
    ...createBuiltinActions({
      integrations,
      catalogClient,
      config: env.config,
      reader: env.reader,
    }),
  ];
  ``` so no plugin change is needed

punkle: I think the discoveryApi should take care of this for you. I dont know the details. But there can and should be an internal url that the backend communicates on. I dont think this would be isolated to this action. There are other examples of backend plugins that communicate with other parts of the app.

I think creating a config reader for the purpose of the action should not be neccessary.

punkle: e.g. the CatalogClient depends on the discovery api to discover the catalog endpoint in order to look up entities in the catalog.

PR #848, opened 2023-02-12
feat: Add optional callback function to prometheus alerts

mimotej:
This PR introduces an option to specify a callback function for `EntityPrometheusAlertCard`. When used, rows in the table become clickable. The callback function has one argument, which is passed from the component and contains JSON data of the row on which the user clicked. Usage looks like this:
```typescript
<EntityPrometheusAlertCard onRowClick={callbackFunction} />
```
And the callback function can look like this:
```typescript
const callbackFunction = (arg: Alerts | undefined) => {
  ...
};
```
Where type `Alerts` is user-defined and is just used to parse JSON easily (also type `any` can be used).

### DIFF
--- a/.changeset/nine-jeans-pump.md
+++ b/.changeset/nine-jeans-pump.md
@@ -1,0 +1,21 @@
+---
+'@roadiehq/backstage-plugin-prometheus': minor
+---
+
+Users can now add a callback to an `EntityPrometheusAlertCard` component. The component with callback can look like this:
+
+```typescript
+<EntityPrometheusAlertCard onRowClick={callbackFunction} />
+```
+
+And `callbackFunction` can have the following definition:
+
+```typescript
+const callbackFunction = (arg: Alerts) => {
+  ...
+};
+```
+
+Where the `Alerts` type is a user-defined type to more easily parse JSON definition (`any` type can also be used). This callback is optional; if not supplied, tables in the row are not clickable.
+
+This change modifies `PrometheusAlertStatus`, which adds `onRowClick` event to a `Table` component.
--- a/plugins/frontend/backstage-plugin-prometheus/README.md
+++ b/plugins/frontend/backstage-plugin-prometheus/README.md
@@ -191,6 +191,25 @@ prometheus:
       proxyPath: /prometheusTeamB/api
 ```
 
+## Using callback with `EntityPrometheusAlertCard`
+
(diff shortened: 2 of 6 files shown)


plugins/frontend/backstage-plugin-prometheus/src/types.ts
```diff
@@ -79,3 +79,5 @@ export type PrometheusRuleResponse = {
   };
   status: 'success' | 'error';
 };
+
+export type OnRowClick = (arg: PrometheusDisplayableAlert | undefined) => void;
```
Xantier, line 83: We'd always have row data in this case right? Otherwise the row itself would be visible/clickable at all.

Xantier (dismissed):
Waiting for callback function type change and the good to merge

plugins/frontend/backstage-plugin-prometheus/README.md
```diff
@@ -191,6 +191,25 @@ prometheus:
...
+The component with callback can look like this:
+
+```typescript
+<EntityPrometheusAlertCard onRowClick={callbackFunction} />
+```
+
+Where `callbackFunction` can have the following definition:
+
+```typescript
+const callbackFunction = (arg: Alerts | undefined) => {
```
Xantier, line 206:
```suggestion
const callbackFunction = (arg: Alerts) => {
```

.changeset/nine-jeans-pump.md
```diff
@@ -0,0 +1,21 @@
...
+Users can now add a callback to an `EntityPrometheusAlertCard` component. The component with callback can look like this:
+
+```typescript
+<EntityPrometheusAlertCard onRowClick={callbackFunction} />
+```
+
+And `callbackFunction` can have the following definition:
+
+```typescript
+const callbackFunction = (arg: Alerts | undefined) => {
```
Xantier, line 14:
```suggestion
const callbackFunction = (arg: Alerts) => {
```

plugins/frontend/backstage-plugin-prometheus/src/types.ts
mimotej: Fixed thank you :+1:

Xantier (approved):

Xantier:
:shipit:

Xantier:
Shipped as part of release just now. Version 2.4.0 should be available in repositories soon.

PR #851, opened 2023-02-13
UI enhancement

### LINKED ISSUE #850: UI elements overlap when referenced entity has long organisation and/or repository names
## Expected Behavior

Organisation & repo name text is truncated or filter area is scrollable.

## Current Behavior

Organisation & repo name text is cover with table's filter area.

Screenshot 2023-02-13 at 14 32 53

## Steps to Reproduce

Follow installation steps in the plugin README. Create repository with long name for adding it to the entity catalog or manually edit HTML in the dev tools to contain longer test.

## Possible Solution

Truncate logs table header text or add possibility to scroll filter area when repository name exceeds maximum length.

screenshot1

screenshot2

## Your Environment

- `@roadiehq/backstage-plugin-github-pull-requests` version: **2.5.0**

frenchbread:
closes #850 

### changes

> Fixes bug in PullRequestsTableView component where table layout did not readjust properly when table header text length exceeded.

### screenshot before

screenshot1

### screenshot after

screenshot2
screenshot3

### DIFF
--- a/plugins/frontend/backstage-plugin-github-pull-requests/src/components/PullRequestsTable/PullRequestsTable.tsx
+++ b/plugins/frontend/backstage-plugin-github-pull-requests/src/components/PullRequestsTable/PullRequestsTable.tsx
@@ -157,13 +157,33 @@ export const PullRequestsTableView: FC<Props> = ({
       onRowsPerPageChange={onChangePageSize}
       title={
         <>
-          <Box display="flex" alignItems="center">
+          <Box
+            position="absolute"
+            left="20px"
+            top="17px"
+            width="45%"
+            display="flex"
+            alignItems="center"
+          >
             <GitHubIcon />
             <Box mr={1} />
-            <Typography variant="h6">{projectName}</Typography>
+            <Typography variant="h6" noWrap>
+              {projectName}
+            </Typography>
           </Box>
-          {StateFilterComponent ? <StateFilterComponent /> : <></>}
-          {SearchComponent ? <SearchComponent /> : <></>}
+          <Box
+            position="absolute"
+            right="20px"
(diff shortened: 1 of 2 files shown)


plugins/frontend/backstage-plugin-github-pull-requests/src/components/PullRequestsTable/PullRequestsTable.tsx
```diff
@@ -137,6 +137,17 @@ export const PullRequestsTableView: FC<Props> = ({
...
+    const elem = document.querySelector('.gh-pull-requests-table-header');
+
+    const targetElement = elem?.parentElement?.parentElement?.parentElement;
+
+    if (targetElement) {
+      targetElement.style.display = 'block';
+      targetElement.style.paddingRight = '24px';
+    }
+  }, []);
+
```
kissmikijr, line 150: I do not really like this solution. Can you please try to do this with CSS only? I think something like setting the overflow on the title element would bring us 90% there.

My other thought is that this <Table/> component is coming from backstage/backstage. I think we should try to fix this title in the base component's handling of the title prop so every backstage table would be able to handle the overflow of big titles

frenchbread: @kissmikijr Thanks for your comment. I've updated PR to make is CSS only. Result looks almost the same.

kissmikijr:
We'll need one more rebase, the build pipeline was broken and the fix should be out now in the main branch

kissmikijr (approved):

kissmikijr:
Thank you, let's merge it!

PR #860, opened 2023-02-20
make sure we run on node 16

kissmikijr:
Pin the node version properly in the github actions.


punkle (approved):

.github/workflows/changeset.yml
```diff
@@ -9,8 +9,17 @@ jobs:
   create-release-pr:
     name: Create Changeset PR
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16]
```
punkle, line 14: Should we also try to support v18?

PR #883, opened 2023-03-13
Remove /api from readme for scaffolder-backend-module-http-request

simonstamm:
`/api` is duplicated in the request since the baseUrl is now being set using the discovery API (calling `discovery.getBaseUrl()`) which adds `/api` to the URL anyway.


kissmikijr (approved):
thank you!

PR #886, opened 2023-03-15
add ability to configure a custom group/user names and hierarchy

punkle:
Add ability to configure a custom group/user names, and provide the name of the parent group to create a hierarchy of groups.

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
@@ -73,6 +73,7 @@ describe('OktaGroupProvider', () => {
               name: 'Everyone@the-company',
               description: 'Everyone in the company',
               org_id: '1234',
+              parent_org_id: '1234',
             },
             listUsers: () => {
               return new MockOktaCollection([
@@ -90,6 +91,8 @@ describe('OktaGroupProvider', () => {
             profile: {
               name: 'Everyone@the-company',
               description: null,
+              org_id: '1235',
+              parent_org_id: '1234',
             },
             listUsers: () => {
               return new MockOktaCollection([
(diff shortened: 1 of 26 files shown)


plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
```diff
@@ -172,5 +177,134 @@ describe('OktaGroupProvider', () => {
...
+      const provider = OktaGroupEntityProvider.fromConfig(config, {
+        logger,
+        namingStrategy: () => {
+          throw new Error('bork');
+        },
+        userNamingStrategy: 'strip-domain-email',
+      });
+      await provider.connect(entityProviderConnection);
+      await provider.run();
+      expect(entityProviderConnection.applyMutation).toBeCalledWith({
```
Xantier, line 271: Should this assert `neverHaveBeenCalled` instead of empty array?

Xantier (dismissed):

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
punkle: I think this is correct. Because the provider has found 0 groups so it is going to emit an empty set. This makes sense when you change a filter where previously it was finding more entities than before.

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.test.ts
```diff
@@ -187,5 +189,217 @@ describe('OktaOrgEntityProvider', () => {
...
+        applyMutation: jest.fn(),
+        refresh: jest.fn(),
+      };
+      const provider = OktaOrgEntityProvider.fromConfig(config, {
+        logger,
+        groupNamingStrategy: new ProfileFieldGroupNamingStrategy('org_id')
+          .nameForGroup,
+        userNamingStrategy: 'strip-domain-email',
+      });
+      provider.connect(entityProviderConnection);
```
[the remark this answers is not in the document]
punkle: yes

PR #908, opened 2023-03-23
Added capacity to parse profile attributes into annotations

tomp21:
This adds the capability of setting custom attributes from a group into annotations on the corresponding backstage entity

In our use case, we need to enrich the group entities to have some of the profile attributes the Okta group has, and i couldn't think of another way than moving them to annotations

May be worth implement the same for user entities as well? I have no problem adding that if this feature is accepted

- [ :heavy_check_mark: ] Added tests for new functionality and regression tests for bug fixes
- [ :heavy_check_mark: ] Added changeset (run `yarn changeset` in the root)
- [ N/A ] Screenshots of before and after attached (for UI changes)
- [ TODO ] Added or updated documentation (if applicable)


plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -103,13 +110,25 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
           this.logger.warn(`failed to add user to group: ${e.message}`);
         }
       });
-
+      const profileAnnotations: Record<string, string> = {};
+      if (this.customAttributesToAnnotationWhitelist.length) {
+        for (const [key, value] of new Map(Object.entries(group.profile))) {
+          const stringKey = key.toString();
+          if (this.customAttributesToAnnotationWhitelist.includes(stringKey)) {
+            profileAnnotations[stringKey] = value.toString();
```
kissmikijr, line 118: Could you please namespace the attribute with something ? For example "okta.com/<stringKey>"

kissmikijr (changes requested):
Thank you for this. Can you please update the readme with the useful examples how to utilize this new feature ?

kissmikijr:
I think it makes sense to add this to the User entities as well. @punkle what do you think about this?

tomp21:
> Contributor

Sure thing! I just wanted to make sure this feature made sense to you too before putting too much time on it

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -103,13 +110,25 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
           this.logger.warn(`failed to add user to group: ${e.message}`);
         }
       });
-
+      const profileAnnotations: Record<string, string> = {};
+      if (this.customAttributesToAnnotationWhitelist.length) {
+        for (const [key, value] of new Map(Object.entries(group.profile))) {
```
punkle, line 115: I wish `GroupEntity['spec']['profile']` was more flexible. I wonder what the tradeoffs are between using labels vs using annotations? Thoughts?

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
```diff
@@ -63,6 +63,73 @@ describe('OktaGroupProvider', () => {
...
+    });
+
+    it('passes the whitelisted fields to annotations', async () => {
+      const entityProviderConnection: EntityProviderConnection = {
+        applyMutation: jest.fn(),
+        refresh: jest.fn(),
+      };
+      const provider = OktaGroupEntityProvider.fromConfig(config, {
+        logger,
+        customAttributesToAnnotationWhitelist: [
```
punkle, line 102: How about `customAttributesToAnnotationAllowlist` instead?

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -64,6 +66,7 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
       parentGroupField?: string;
       namingStrategy?: GroupNamingStrategies | GroupNamingStrategy;
       userNamingStrategy?: UserNamingStrategies | UserNamingStrategy;
+      customAttributesToAnnotationWhitelist?: string[];
```
punkle, line 69: Could you give an example of a profile field that you would not want to be loaded into backstage? It feels like the default should be to load them all in? Thoughts?

punkle:
A couple of minor comments and discussion points, but in general I think its a good idea to load in the profile fields. I wonder should we approach the core backstage team to ask them to allow the GroupEntity profile to include custom fields.

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
kissmikijr: To me it makes sense to be an annotation. I think this is exactly what annotations are used for to store some additional data in a key-value pairs that you'll later work on, base some conditional rendering or using them in processors.

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
tomp21: i wanted to keep the previous behaviour as the default, and some organizations might not even need this, and adding everything as an annotation may generate too much clutter, i wanted to be minimalistic by default, we could have an option to just importa all as well

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -103,13 +110,25 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
           this.logger.warn(`failed to add user to group: ${e.message}`);
         }
       });
-
+      const profileAnnotations: Record<string, string> = {};
+      if (this.customAttributesToAnnotationWhitelist.length) {
+        for (const [key, value] of new Map(Object.entries(group.profile))) {
+          const stringKey = key.toString();
+          if (this.customAttributesToAnnotationWhitelist.includes(stringKey)) {
```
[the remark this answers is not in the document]
tomp21: That's something i thought weird, do we really have a case where the `key` of a custom attribute in Okta is an array?

tomp21:
> A couple of minor comments and discussion points, but in general I think its a good idea to load in the profile fields. I wonder should we approach the core backstage team to ask them to allow the GroupEntity profile to include custom fields.

This was my first approach, but after implementing it i think it makes a bunch more sense to map this kind of info to annotations, since this is very variable and could make deserialization more complex than it needs to be

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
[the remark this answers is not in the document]
punkle: These are the types it allows:

image

punkle:
> could make deserialization more complex than it needs to be

What do you mean?

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
tomp21: sure

.changeset/strange-needles-hang.md
```diff
@@ -0,0 +1,5 @@
+---
```
punkle, line 1: Could you also add this feature to `OktaOrgEntityProvider`

tomp21:
> > could make deserialization more complex than it needs to be
> 
> What do you mean?

I mean that adding a dynamic structure as a map[string]string adds complexity to the entity itself, being that it already has the annotations to store this kind of information

tomp21:
I'll be implementing the changes from the comments soon, i just didn't have the time these days to do it

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
tomp21: mmm wouldn't organisations benefit from having their own namespacing there? in the end the sense of this is to be able to have further custom integrations 🤔  i'm open to discuss it though

tomp21:
> * add the feature to the org provider for consistency

shouldn't i add it also for the user provider?

punkle:
> shouldn't i add it also for the user provider?

no harm

amatas-stuart:
Hi @punkle, could you give another review pass to this PR? 

if you think that anything is missing please let us know, thank you :pray:

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -171,14 +178,30 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
               group,
               oktaGroups,
             });
-
+            const profileAnnotations: Record<string, string> = {};
+            if (this.customAttributesToAnnotationAllowlist.length) {
```
punkle, line 182: There is no need to check the length here, you are already looping over the list. Not checking should be equivilent.

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
```diff
@@ -80,9 +87,22 @@ export class OktaUserEntityProvider extends OktaEntityProvider {
     const allUsers = await client.listUsers({ search: this.userFilter });
 
     await allUsers.each(user => {
+      const profileAnnotations: Record<string, string> = {};
```
punkle, line 90: We can refactor this into a function and reuse the function in three places.

e.g.

Can we refactor this into a function to reduce the triplication.

I think the function could take a user/group. e.g.

```typescript
export getCustomAnnotations(member: User | Group, allowList: string[]) {
  const profileAnnotations: Record<string, string> = {};
  for (const [key, value] of new Map(Object.entries(member.profile))) {
    const stringKey = key.toString();
    if (this.customAttributesToAnnotationAllowlist.includes(stringKey)) {
      profileAnnotations[stringKey] = value.toString();
    }
  }
  return profileAnnotations;
}
```

Then you can just do the following:

```typescript
      const annotations = {
        ...defaultAnnotations,
        ...(getCustomAnnotations(user)),
      };
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -136,14 +143,28 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
...
+          ...profileAnnotations,
+        };
         try {
           const groupEntity = groupEntityFromOktaGroup(
             group,
             this.namingStrategy,
             parentGroup,
             {
-              annotations: defaultAnnotations,
+              annotations: annotations,
```
kissmikijr, line 167:
```suggestion
              annotations,
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
```diff
@@ -80,9 +87,22 @@ export class OktaUserEntityProvider extends OktaEntityProvider {
...
+        }
+      }
+      const annotations = {
+        ...defaultAnnotations,
+        ...profileAnnotations,
+      };
       try {
         const userEntity = userEntityFromOktaUser(user, this.namingStrategy, {
-          annotations: defaultAnnotations,
+          annotations: annotations,
```
kissmikijr, line 105:
```suggestion
          annotations,
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -171,14 +178,30 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
...
+              ...profileAnnotations,
+            };
             try {
               const groupEntity = groupEntityFromOktaGroup(
                 group,
                 this.groupNamingStrategy,
                 parentGroup,
                 {
-                  annotations: defaultAnnotations,
+                  annotations: annotations,
```
kissmikijr, line 204:
```suggestion
                  annotations,
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -136,14 +143,28 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
           group,
           oktaGroups,
         });
-
+        const profileAnnotations: Record<string, string> = {};
+        if (this.customAttributesToAnnotationAllowlist.length) {
+          for (const [key, value] of new Map(Object.entries(group.profile))) {
+            const stringKey = key.toString();
```
kissmikijr, line 149: I'd rather not have its type in the variable name. It should be obvious it is a string after the toString().

We could use let and re-use the key variable.

punkle: I actually dont think that this line is necessary. The key should always be a string. https://github.com/okta/okta-sdk-nodejs/blob/okta-sdk-nodejs-6.6.0/src/types/models/GroupProfile.d.ts#L27 same with UserProfile.

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -136,14 +143,28 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
...
+            if (
+              this.customAttributesToAnnotationAllowlist.includes(stringKey)
+            ) {
+              profileAnnotations[stringKey] = value.toString();
```
punkle, line 153: I think how we serialize this value could be better.. It can be any of these types:

```boolean | number | string | boolean[] | number[] | string[]```

Maybe we could change this to:

```
profileAnnotations[stringKey] = JSON.stringify(value);
```

Comment (punkle) [reply plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts:153]: Here is some examples. See the true value as a string is undecernable from the the boolean true using toString, but it is correct using JSON.stringify.

```
> true.toString()
'true'
> "true".toString()
'true'
> JSON.stringify("true")
'"true"'
> JSON.stringify(true)
'true'
```

**Comment by tomp21** [comment]
Sorry i took a couple weeks AFK, i'm getting back to it with @zinahia 's help

File: plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts (hunk 12 above)
Comment (tomp21) [reply plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts:149]: mmm i was pretty sure i did this for a reason but removing it an rerunning the tests just works so i guess i got confused at the moment i wrote this, i'll remove this line

**Comment by zinahia** [comment]
Hey @punkle I think we have applied all of your suggestions except for [this one](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/908#discussion_r1168629929) we're a bit torn about it because it would mean that the custom annotations would have a different format in the value compared to the rest which are simple string values, instead it would have a JSON value wrapped inside the string. It would end up looking like this:
```
"metadata": Object {
  "annotations": Object {
    "backstage.io/managed-by-location": "okta-user-https://okta",
    "backstage.io/managed-by-origin-location": "okta-user-https://okta",
    "customAttribute1": "\"asdfasdf\"",
    ...
}
```

I wonder if it's worth doing this wrapping to handle the boolean differentiation if the clients could handle the transformation on their side if they needed to. Personally I think it's a bit cleaner to leave the `toString()` and be consistent with the rest of the annotations but we're open to changing it if you prefer it that way.

**Comment by punkle** [comment]
@zinahia I think you have me convinced.

I had a look at k8s annotations and I think that the annotations only really support something like "Record<string, string>". So it is a bit of a challenge trying to represent object, array or boolean in here. It is unfortunate, but I cant think of a better way to represent these things without using toString as you have done.

**Review by punkle — APPROVED** [review]

File: plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts [hunk 14]
```diff
@@ -52,13 +54,18 @@ export class OktaUserEntityProvider extends OktaEntityProvider {
...
     },
   ) {
     super([accountConfig], options);
     this.namingStrategy = userNamingStrategyFactory(options.namingStrategy);
     this.userFilter = accountConfig.userFilter;
     this.orgUrl = accountConfig.orgUrl;
+    this.customAttributesToAnnotationAllowlist =
+      options.customAttributesToAnnotationAllowlist
+        ? options.customAttributesToAnnotationAllowlist
+        : [];
```
kissmikijr, line 68: we could write this expression like the below suggestion. I think it is easier to read.
```suggestion
    this.customAttributesToAnnotationAllowlist = options.customAttributesToAnnotationAllowlist || [];
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.test.ts
```diff
@@ -194,5 +195,34 @@ describe('OktaUserEntityProvider', () => {
...
+      await provider.run();
+      expect(entityProviderConnection.applyMutation).toBeCalledWith({
+        type: 'full',
+        entities: [
+          expect.objectContaining({
+            entity: expect.objectContaining({
+              kind: 'User',
+              metadata: expect.objectContaining({
+                annotations: expect.objectContaining({
+                  customAttribute1: 'asdfasdf',
```
kissmikijr, line 218: Would you mind using some kind of real word instead of these random chars?

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.test.ts
```diff
@@ -194,5 +195,34 @@ describe('OktaUserEntityProvider', () => {
...
+                }),
+                name: 'asdfwefwefwef',
```
kissmikijr, line 220: Same here please

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -91,6 +94,10 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
     );
     this.orgUrl = accountConfig.orgUrl;
     this.groupFilter = accountConfig.groupFilter;
+    this.customAttributesToAnnotationAllowlist =
+      options.customAttributesToAnnotationAllowlist
+        ? options.customAttributesToAnnotationAllowlist
+        : [];
```
kissmikijr, line 100: we could write this expression like the below suggestion. I think it is easier to read.
```suggestion
    this.customAttributesToAnnotationAllowlist = options.customAttributesToAnnotationAllowlist || [];
```

plugins/backend/catalog-backend-module-okta/src/providers/OktaUserEntityProvider.ts
zinahia: Done in: https://github.com/RoadieHQ/roadie-backstage-plugins/pull/908/commits/22d7bd33793c4fd5aab31bdde831c175701ca2d0

kissmikijr (approved):
Really good work, thank you for the contribution!

PR #910, opened 2023-03-24
Export `EntityDependabotAlertsTable` and `EntitySecurityInsightsTable` from security insights plugin

dotboris:
This PR takes the existing `DependabotAlertsTable` and `SecurityInsightsTable` components from the security insights plugin and exports them. This will allow users to build their own views by using the tables directly.

I have a use case for this. I want to have a "Security" tab that holds both information about the dependabot alerts and the security insights. This will allow me to do that.

### DIFF
--- a/plugins/frontend/backstage-plugin-security-insights/src/components/SecurityInsightsTable/SecurityInsightsTable.tsx
+++ b/plugins/frontend/backstage-plugin-security-insights/src/components/SecurityInsightsTable/SecurityInsightsTable.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import React, { FC, useState } from 'react';
+import React, { useState } from 'react';
 import { Typography, Box, Grid } from '@material-ui/core';
 import GitHubIcon from '@material-ui/icons/GitHub';
 import { useApi, githubAuthApiRef } from '@backstage/core-plugin-api';
@@ -29,25 +29,21 @@ import { UpdateSeverityStatusModal } from '../UpdateSeverityStatusModal';
 import { StateFilterComponent } from './components/StateFilterComponent';
 import { useUrl } from '../useUrl';
 import { getSeverityBadge } from '../utils';
-import {
-  SecurityInsightsTabProps,
-  SecurityInsight,
-  SecurityInsightFilterState,
-} from '../../types';
+import { SecurityInsight, SecurityInsightFilterState } from '../../types';
+import { useEntity } from '@backstage/plugin-catalog-react';
 
 const getElapsedTime = (start: string) => {
   return moment(start).fromNow();
 };
 
-export const SecurityInsightsTable: FC<SecurityInsightsTabProps> = ({
-  entity,
-}) => {
+export const SecurityInsightsTable = () => {
   const [insightsStatusFilter, setInsightsStatusFilter] =
     useState<SecurityInsightFilterState>(null);
   const [filteredTableData, setFilteredTableData] = useState<SecurityInsight[]>(
     [],
   );
   const [tableData, setTableData] = useState<SecurityInsight[]>([]);
+  const { entity } = useEntity();
   const { owner, repo } = useProjectEntity(entity);
   const projectName = useProjectName(entity);
   const auth = useApi(githubAuthApiRef);
--- a/.changeset/serious-points-cry.md
+++ b/.changeset/serious-points-cry.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-security-insights': minor
+---
+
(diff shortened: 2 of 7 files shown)


plugins/frontend/backstage-plugin-security-insights/src/components/SecurityInsightsTable/SecurityInsightsTable.tsx
```diff
@@ -29,25 +29,21 @@ import { UpdateSeverityStatusModal } from '../UpdateSeverityStatusModal';
...
-  entity,
-}) => {
+export const SecurityInsightsTable = () => {
   const [insightsStatusFilter, setInsightsStatusFilter] =
     useState<SecurityInsightFilterState>(null);
   const [filteredTableData, setFilteredTableData] = useState<SecurityInsight[]>(
     [],
   );
   const [tableData, setTableData] = useState<SecurityInsight[]>([]);
+  const { entity } = useEntity();
```
kissmikijr, line 46: Thank you for the refactor !

kissmikijr (commented):
I'm thinking why do we need to provide the individual components from this plugin ? Couldn't we just export them from this package and you could build your own out of them then?

dotboris:
> I'm thinking why do we need to provide the individual components from this plugin ? Couldn't we just export them from this package and you could build your own out of them then?

I'm not sure I understand what you mean here. You're asking why we should "provide" the components and why we couldn't just "export" them. I'm confused because to me "export" and "provide" effectively mean the same thing.

dotboris:
I thought about this a little more and I think that I understand. Are you asked why I expose the components through `securityInsightsPlugin.provide` instead of just directly `export`ing them?

If so, my understanding was that this was just the way plugins were meant to expose components so that they can be used by the app package. Looking at https://backstage.io/docs/plugins/composability/#extensions, it seems like it's what backstage recommends plugins do.

kissmikijr:
> I thought about this a little more and I think that I understand. Are you asked why I expose the components through securityInsightsPlugin.provide instead of just directly exporting them?

Yes I meant exactly this, sorry!: ) 

> If so, my understanding was that this was just the way plugins were meant to expose components so that they can be used by the app package. Looking at https://backstage.io/docs/plugins/composability/#extensions, it seems like it's what backstage recommends plugins do.

Yea however the plugins usually are not a stand-alone components, usually, they are pages, tabs, and widgets which are built from the components. In your usecase I assume you'd like to reuse our component to build your own version of the widget/page. If this is the case you'd use this package as a UI component library it has nothing to do with backstage from your POV because in the end, you'd create your own plugin which would plugin.provide() your own component which uses these components. Am I correct in thinking this is the case?

dotboris:
> Yea however the plugins usually are not a stand-alone components, usually, they are pages, tabs, and widgets which are built from the components. In your usecase I assume you'd like to reuse our component to build your own version of the widget/page. If this is the case you'd use this package as a UI component library it has nothing to do with backstage from your POV because in the end, you'd create your own plugin which would plugin.provide() your own component which uses these components. Am I correct in thinking this is the case?

In my use case, I plan on adding a tab to the component page and put both of the tables in there. I'm not creating an extra plugin. It'll all be done directly within the app. In that sense I feel that I'm using this plugin as a plugin. I think that these two tables fit the general definition of what plugins normally provide. At least, I think that it matches what the backstage doc says:

>Component extensions are plain React component with no particular requirements, for example a card for an entity overview page. The component will be exported more or less as is, but is wrapped to provide things like an error boundary, lazy loading, and a plugin context.
>
> https://backstage.io/docs/plugins/composability/#extensions

The way I see it is that it's normal for plugins to provide fully featured components that users can drop into their catalog pages. In this case, I think that these two tables fit the bill. Like the other widgets that are exposed by this plugin, they can be dropped in anywhere within the app / catalog pages and just work. The only real difference is that the existing widgets are cards while these are tables.

That being said, I'm not particularly attached to using `plugin.provide(...)`. If it's not done here I'm probably going to have to do it myself on my end just so that I can benefit from the error boundary and lazy loading. If you'd like me to, I can change this to be a direct export.

kissmikijr:
> The way I see it is that it's normal for plugins to provide fully featured components. 

I disagree, fully featured components are provided by regular UI libraries. I don't think these tables make sense on their own. If some new adopters would come and they see this plugin they could naively end up adding the tables only to their instance without any pointers what is it. No title, no page, no support helper icons etc. 

I'm not against the idea to reuse these components and let adopters build their own widgets on top of them. So I suggest we go with a regular export so people can use this as a package to build their version of the widgets.

kissmikijr (approved):
Thank you!

PR #915, opened 2023-03-28
add lineWidth option to yaml.dump for JSONata and serialization actions

rmartine-ias:
The JSONata action is very useful for modifying an existing YAML file. However, it fails to match the style of the destination YAML, creating large diffs. If you have a file that looks like this:

```yaml
someKey: moreThan80Chars
```

and run a transformation on it (`$ ~> | $ | { "newKey": "newVal" } |`), the end result is something like:

```yaml
someKey: >-
  moreThan80Chars
newKey: newval
```

For a file with many such lines (in our case, AWS ARNs), this creates a massive diff against the file, when only a small transformation was performed.

This change allows for setting the `lineWidth` property of [`js-yaml.dump`][0], which when set to `-1`, disables auto-wrapping. This would allow us to use JSONata transformations without enforcing the default `js-yaml` style on the destination YAML.

#### Looking for Feedback On

1. I wasn't sure which style of test is preferred -- testing that the option was passed to `yaml.dump`, or testing that the output actually didn't wrap. I left in one of each style, and can change the other to the preferred, if this change is something you would like merged.
   1. If "test expected output", should there be a negative test, that without the option passed, line wrapping is performed?
1. The JSONata transformation is very simple, replacing the (empty) file wholesale. Should this be more complex?
1. Should I add the other relevant `dump` options while I'm in here? `noArrayIndent`, `skipInvalid`, `flowLevel`, `sortKeys`, `noRefs`, `noCompatMode`, `condenseFlow`, `quotingType`, and `forceQuotes` all seem potentially useful.
    1. Should I add them to `utils:merge` as well?
    1. I'm new to Typescript, but would this be a case for `pick`/`omit` on the `yaml.dump` parameters? Or would that be too complicated, because each option needs to be specified in the JSON Schema as well?

<details>
<summary>Test in a real template</summary>

With this template:

```yaml
apiVersion: scaffolder.backstage.io/v1beta3
kind: Template

metadata:
  name: test-template
  title: Test template

spec:
  owner: myTeam
  type: service

  parameters:
    - title: Test
      properties:
        test:
          title: Test
          type: string

  steps:
    - id: fetch-base
      name: Fetch Base
      action: fetch:template
      input:
        url: ./skeleton

    - id: update-base
      name: Update Base
      action: roadiehq:utils:jsonata:yaml:transform
      input:
        path: test.yaml
        expression: |-
          $ ~> | $ | {
            "some_super_long_key": "some super super super super super super super super super super super super long value"
          } |

    - id: log-message
      name: Log Message
      action: debug:log
      input:
        message: ${{ steps['update-base'].output.result }}
```

and `skeleton/test.yaml` containing `hello: world`, I ran the template with the suggested plugin changes in a backstage instance, and got this (expected, unwanted result):

```console
12023-03-28T00:37:47.364Z Beginning step Log Message
22023-03-28T00:37:47.367Z info: {
3  "message": "hello: world\nsome_super_long_key: >-\n  some super super super super super super super super super super super super\n  long value\n"
4}
52023-03-28T00:37:47.367Z hello: world
6some_super_long_key: >-
7  some super super super super super super super super super super super super
8  long value
92023-03-28T00:37:47.368Z Finished step Log Message
```

(See the `>-`.) I added `options.lineWidth: -1` to the JSONata step and got this (expected, desired result):

```console
12023-03-28T00:41:15.878Z Beginning step Log Message
22023-03-28T00:41:15.882Z info: {
3  "message": "hello: world\nsome_super_long_key: some super super super super super super super super super super super super long value\n"
4}
52023-03-28T00:41:15.885Z hello: world
6some_super_long_key: some super super super super super super super super super super super super long value
72023-03-28T00:41:15.886Z Finished step Log Message
```
</details>

[0]: https://www.npmjs.com/package/js-yaml

   - N/A, no UI changes
  - no existing docs for `indent` option; assuming auto-generated backstage `/create/action` docs are OK

### DIFF
--- a/.changeset/late-ducks-heal.md
+++ b/.changeset/late-ducks-heal.md
@@ -2,4 +2,4 @@
 '@roadiehq/scaffolder-backend-module-utils': patch
 ---
 
-Added lineLength option to serialize:yaml and jsonata actions to allow for control of line wrapping
+Added most yaml.dump options to serialize, merge, and jsonata actions
--- a/.changeset/late-ducks-heal.md
+++ b/.changeset/late-ducks-heal.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
+---
+
+Added lineLength option to serialize:yaml and jsonata actions to allow for control of line wrapping
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.test.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.test.ts
@@ -55,4 +55,30 @@ describe('roadiehq:utils:jsonata:yaml:transform', () => {
       yaml.dump({ hello: ['world', 'item2'] }),
     );
   });
+
+  it('should use lineWidth to control wrapping lines', async () => {
+    mock({
+      'fake-tmp-dir': {
+        'fake-file.yaml': '',
+      },
+    });
+    await action.handler({
+      ...mockContext,
+      workspacePath: 'fake-tmp-dir',
+      input: {
+        path: 'fake-file.yaml',
+        expression:
+          '{ "helloFrom": "anIncrediblyLongEmailAddress@anIncrediblyLongSubdomain.anIncrediblyLongDomain.com" }',
+        options: {
+          lineWidth: -1,
+        },
+      },
+    });
+
+    // No block style ( >- )
+    expect(mockContext.output).toHaveBeenCalledWith(
+      'result',
+      'helloFrom: anIncrediblyLongEmailAddress@anIncrediblyLongSubdomain.anIncrediblyLongDomain.com\n',
+    );
+  });
 });
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.test.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.test.ts
@@ -56,29 +56,39 @@ describe('roadiehq:utils:jsonata:yaml:transform', () => {
     );
   });
 
-  it('should use lineWidth to control wrapping lines', async () => {
+  it('should pass options to yaml.dump', async () => {
     mock({
       'fake-tmp-dir': {
         'fake-file.yaml': '',
       },
     });
+
+    const mockDump = jest.spyOn(yaml, 'dump');
+
+    const opts = {
+      indent: 3,
+      noArrayIndent: true,
+      skipInvalid: true,
+      flowLevel: 23,
+      sortKeys: true,
+      lineWidth: -1,
+      noRefs: true,
+      noCompatMode: true,
+      condenseFlow: true,
+      quotingType: '"' as const,
+      forceQuotes: true,
+    };
+
     await action.handler({
       ...mockContext,
       workspacePath: 'fake-tmp-dir',
       input: {
         path: 'fake-file.yaml',
-        expression:
-          '{ "helloFrom": "anIncrediblyLongEmailAddress@anIncrediblyLongSubdomain.anIncrediblyLongDomain.com" }',
-        options: {
-          lineWidth: -1,
-        },
+        expression: '{ "hello": "beautiful world" }',
+        options: opts,
       },
     });
 
-    // No block style ( >- )
-    expect(mockContext.output).toHaveBeenCalledWith(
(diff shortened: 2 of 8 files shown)


kissmikijr:
Thank you for this PR! Kudos for the really well-written PR description! :) 

> I wasn't sure which style of test is preferred -- testing that the option was passed to yaml.dump, or testing that the output actually didn't wrap. I left one of each style and can change the other to the preferred if this change is something you would like merged.
If "test expected output", should there be a negative test, that without the option passed, line wrapping is performed?

I think in this case it is enough to test that we pass the options down to the dump. I don't believe we should test the underlying package's behaviour in this code.

> The JSONata transformation is very simple, replacing the (empty) file wholesale. Should this be more complex?

Sorry, which transformation do you refer to here? 

> Should I add the other relevant dump options while I'm in here? noArrayIndent, skipInvalid, flowLevel, sortKeys, noRefs, noCompatMode, condenseFlow, quotingType, and forceQuotes all seem potentially useful.

It would be appreciated as these make sense! But don't feel pressured. This PR can be a really good example if someone needs their options should be easy to add when they are needed! :) 

> Should I add them to utils:merge as well?

Let's add this to the utils:merge as well.

> I'm new to Typescript, but would this be a case for pick/omit on the yaml.dump parameters? Or would that be too complicated, because each option needs to be specified in the JSON Schema as well? 

I think you are correct, because we need to specify them in the JSON schema we wouldn't win that much, but I think we might use the yaml.dump's params' type in the function parameter.

rmartine-ias:
> Thank you for this PR! Kudos for the really well-written PR description! :)

Thank you!

> I think in this case it is enough to test that we pass the options down to the dump. I don't believe we should test the underlying package's behaviour in this code.

Makes sense to me, will update with this.

> Sorry, which transformation do you refer to here?

The `expression` on line 70 in the JSONata test:

```js
       expression:
          '{ "helloFrom": "anIncrediblyLongEmailAddress@anIncrediblyLongSubdomain.anIncrediblyLongDomain.com" }',
```

I think it's fine to leave since the test is whether the options are passed.

> It would be appreciated as these make sense! But don't feel pressured. This PR can be a really good example if someone needs their options should be easy to add when they are needed! :)
> Let's add this to the utils:merge as well.

Will do.

> I think you are correct, because we need to specify them in the JSON schema we wouldn't win that much, but I think we might use the yaml.dump's params' type in the function parameter.

I'll see if I can get this working nicely, but won't spend too much time on it.

rmartine-ias:
@kissmikijr updated to include other options, and added to utils:merge. Ready for review.

As I was looking at the tests I thought "hang on, this is testing the *result* of `yaml.dump` with the given options. We'd need to find a set of options and input where each option changes something about the output. Otherwise this test could pass, while the underlying code fails to pass an option." So I mocked out the `yaml.dump` call to directly verify the passing of options, instead. The test code is slightly duplicated across the three tests, but I think that is an okay tradeoff to prevent more complexity. I really like how concise these plugins are!

I factored out the piece of the JSON schema for YAML options. I'm not sure if I put the file in the appropriate place (`src/types.ts`, a new file), or factored out the schema piece at the right level. The two other ways I was thinking were including the wrapping `{options: {...}}`, or scoping down to the inside of `properties`. (So like `{indent: {...}, noArrayIndent: {...}}`.)  I didn't spend too long thinking about future-proofing this because I think that refactor can come if and when it is needed, but if you think a different way is better, I am happy to change it.

I used the same `options` key for utils:merge, even though it supports JSON as well. My reasoning is that the only option available to `JSON.stringify` is the indent, which `options.indent` covers if anyone needs to add that in the future. This way there is less complexity for changing configuration if you change a step (so not needing to change `options` to `yamlOptions` or something if you change from `serialize` to `merge`).

On another note, how cool would it be if the `/create/actions` page had recursive input properties display capabilities... I added descriptions to the options for in case it ever does. Until then, they might still save people a hop in finding out what they mean. This is a potential maintenance burden though, so I kept it as a separate commit in case you would like me to drop it.

kissmikijr:
Thank you! Let's ship it!

> On another note, how cool would it be if the /create/actions page had recursive input properties display capabilities... I added descriptions to the options for in case it ever does. Until then, they might still save people a hop in finding out what they mean. This is a potential maintenance burden though, so I kept it as a separate commit in case you would like me to drop it.

Yea would be really nice, I hope it will come sometime to the scaffolder! :)

kissmikijr (approved):

rmartine-ias:
Sweet! Running in production for us, now.

PR #921, opened 2023-03-31
add smarter pruning of empty groups

punkle:
add smarter pruning of empty groups

Previously, parent groups of groups with members were being pruned. This change includes parent groups if their decendants have members.

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
@@ -35,6 +35,7 @@ import { getAccountConfig } from './accountConfig';
 import { assertError } from '@backstage/errors';
 import { getOktaGroups } from './getOktaGroups';
 import { getParentGroup } from './getParentGroup';
+import { GroupTree } from './GroupTree';
 
 /**
  * Provides entities from Okta Org service.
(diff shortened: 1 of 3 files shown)


plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -124,14 +126,22 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
...
+            const userName = this.userNamingStrategy(user);
+            userResources[userName] = userEntityFromOktaUser(
+              user,
+              this.userNamingStrategy,
+              {
+                annotations: defaultAnnotations,
+              },
+            );
+          } catch (e: unknown) {
+            assertError(e);
```
sblausten, line 139: Are you sure you want to use this here - this will throw an exception and exit the function just because you can't read the message property to log it, where it seems you want the error to be swallowed.

kissmikijr (approved):

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
sblausten: You are breaking the execution of the rest of the code - is that what you want to do? Its preferable to me to log something even if the error message isn't there than log nothing and throw an exception up the stack that just says that the shape of something was not an error.. Does that make sense?

sblausten: We could/should implement an error log function that takes an error as the standard console.error does as a second arg, and then deals with doing the best job it can at logging something useful, whether thats stringifying the object or accessing the message and stack trace.

sblausten: i.e. Screenshot 2023-03-31 at 16 47 57

sblausten: Throwing a generic type exception from code that is supposed to improve error handling seems like throwing the baby out with the bathwater to me

Issue #935, opened 2023-04-06
Add a href to revision in history table


arnouthoebreckx:
## Feature Suggestion

Would like to add a href to the revision in the History Panel, so that it's easier for developers to be directly redirected to the exact commit that is deployed on ArgoCD. 
ArgoCD has this future for themselves aswell as seen in this picture.

![image](https://user-images.githubusercontent.com/24568583/230483103-50a12e82-4540-4c7e-b13c-d0ba4318a6fc.png)

## Possible Implementation

I would 'copy' the code that ArgoCD uses to generate this URL: https://github.com/argoproj/argo-cd/blob/master/ui/src/app/shared/components/urls.ts

## Context

In general I want to provide an easy way for people to go back into the Backstage history and open up the different deploys that occurred in Git. The change is not very disruptive but can be a quick and easy addition.

I'm creating the contribution at the moment :)

PR #936, opened 2023-04-07
Add href link to revision in history

arnouthoebreckx:

- [ X ] Added tests for new functionality and regression tests for bug fixes
- [ / ] Added changeset (run `yarn changeset` in the root)
- [ / ] Screenshots of before and after attached (for UI changes)
- [ X ] Added or updated documentation (if applicable)

Issue #935 is based on the implementation of ArgoCD itself https://github.com/argoproj/argo-cd/blob/master/ui/src/app/shared/components/urls.ts

Before

image

After

image


plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDHistoryCard.tsx
```diff
@@ -48,6 +49,47 @@ const HistoryTable = ({
   const supportsMultipleArgoInstances: boolean = Boolean(
     configApi.getOptionalConfigArray('argocd.appLocatorMethods')?.length,
   );
+  const protocol = (proto: string): string => {
```
kissmikijr, line 52:
```suggestion
  const protocol = (protocol: string): string => {
```
I think we do not spare much with this shortening.

arnouthoebreckx:
Updated 👍

kissmikijr (approved):

kissmikijr:
One last rebase and it should be good to go!

PR #951, opened 2023-04-17
Add support for custom annotation validation schema

rsmarathe:
Add an option using custom annotation schema for validation. 

In our use case, we have custom annotations. We would like to use [Backstage-entity-validator](https://github.com/RoadieHQ/backstage-entity-validator) GitHub action for validation. But it doesn't provide the option to add custom annotation validation schema. So proposing this solution. 

Users can add endpoints to their Backstage instance backend for exposing annotation schema with custom rules. That URL can be passed from GitHub action to `validate` function here.  If a custom schema is provided, then `validate` can use instead of the hardcoded one here. 


kissmikijr:
Thank you for the contribution. Just a heads up. It might take a while to asses this feature.

rsmarathe:
@kissmikijr Yes, that makes sense.  Switched to reading from the file system. Opened PR for passing location from GitHub action too https://github.com/RoadieHQ/backstage-entity-validator/pull/63.

.changeset/late-cobras-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/roadie-backstage-entity-validator': minor
+---
+
+Add option to provide url for fetching custom annotation schema for validation
```
kissmikijr, line 5:
```suggestion
Add option to specify location of the schema file.
```

kissmikijr (approved):

PR #981, opened 2023-05-13
ArgoCD: use argocd instance url for dashboard url

dweber019:
This will use the instance url for the dashboard link if the `argocd.baseUrl` isn't defined.
We need this as we have multiple instances of argocd, all with own dashboards.


kissmikijr (approved):
Thanks, it looks good to me. @karlhaworth anything to add?

Issue #996, opened 2023-05-24
[backstage-plugin-jira] Hise issue filter by configuratin


stephane-mori:
Hello,

I installed the jira plugin to test it, and I met problem with the issue filter. The jira workflow configuration is very complicated and states on this filter are hard to apprehend. That's why I want to provide a patch hide this filter by configuration. Do you think this patch could be accepted and merged for the next release ?

Regards,
Stéphane Mori

stephane-mori:
Linked to pull request : https://github.com/RoadieHQ/roadie-backstage-plugins/pull/997

stephane-mori:
How can I link the pull request to this issue ? I can see "No branches or pull requests" in right menu.

PR #997, opened 2023-05-24
Allow to hide issue filter by configuration

stephane-mori:


stephane-mori:
Linked to issue : https://github.com/RoadieHQ/roadie-backstage-plugins/issues/996

kissmikijr:
Can you please provide  before|after pictures?

stephane-mori:
after
before

kissmikijr:
I'm a bit puzzled by this introducing this would start a pattern where potentially we could be asked why we decided to support some kind of frontend rendering based on the app-config configuration which I'm not a big fan of. Can you please elaborate on why are you trying to remove this? Do you have a use case where people got confused by how it works? Tbh I don't think it hurts that it is there it's not required to use it.

stephane-mori:
The linked jira workflow configuration is very complicated and states on this filter are hard to apprehend. That's why I want to provide a patch hide this filter by configuration.

Do you prefer an other way than app-config configuration to configure this need ?
I want this filter dissapear. I prefer configuration way, but if isn't possible, i will have to fork.

kissmikijr:
My problem is the configuration via the app-config.yaml. I'd prefer if this would be a regular prop setting on the JiraCard so when you implement it in your backstage you could just say <JiraCard filter={false}/>

stephane-mori:
I don't know how to add a commit to this PR.
I just created another PR .... https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1035

stephane-mori:
Done !

https://github.com/RoadieHQ/roadie-backstage-plugins/pull/997/commits/3e9dab9d83af70f566b5c3d6d6eef7e1c95f3be1

kissmikijr (approved):
Thank you! This is exactly what I was thinking!

Issue #999, opened 2023-05-25
Update the maximum node version to node 18


schultzp2020:
## Feature Suggestion

Update the maximum node version to node 18

## Possible Implementation

Update the root level `package.json` to allow node 18. Add node 18 to the [ci](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/.github/workflows/e2e-tests.yml#L65) version matix.

I would be willing to work on this pr.

## Context

Node 18 is the latest LTS version

PR #1002, opened 2023-05-30
chore(monorepo): update node to v18

schultzp2020:
Update node to v18

Fixes https://github.com/RoadieHQ/roadie-backstage-plugins/issues/999

### DIFF
--- a/.github/workflows/changeset.yml
+++ b/.github/workflows/changeset.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16]
+        node-version: [18]
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16]
+        node-version: [14, 16, 18]
     steps:
       - uses: actions/checkout@v3
         with:
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -7,9 +7,6 @@
   "backstage": {
     "role": "backend"
   },
-  "engines": {
-    "node": "14 || 16"
-  },
   "scripts": {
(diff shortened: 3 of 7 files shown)


.github/workflows/changeset.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16]
+        node-version: [18]
```
punkle, line 14: can we also test on 16:

i.e.

```
node-version: [16, 18]
```

.github/workflows/main.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16]
+        node-version: [14, 16, 18]
```
punkle, line 15: why add 14 here?

packages/backend/package.json
```diff
@@ -7,9 +7,6 @@
   "backstage": {
     "role": "backend"
   },
-  "engines": {
```
punkle, line 10: should this include the engines that we are testing with? i.e. 16, 18

.github/workflows/main.yml
schultzp2020: I added 14 because its considered a [supported version](https://github.com/schultzp2020/roadie-backstage-plugins/blob/update-node/package.json#L6). I can remove 14 if you'd like.

.github/workflows/changeset.yml
schultzp2020: I can add 16.

packages/backend/package.json
punkle: ah I see.

.github/workflows/main.yml
punkle: That makes sense now.. thanks.

punkle:
@schultzp2020 would you mind taking a look at those failing tests?

punkle:
Hi @schultzp2020 is there anything we can do to help you with this one?

PR #1037, opened 2023-07-10
POC ability to return object from select field from api

punkle:
POC ability to return object from select field from api

I dont know if this even works, I havent tried to run it..

The select field from api field currently validates that the field returns a string with specific characteristics. The idea with this change is to allow any type to be returned from the api to be presented to the scaffolder action steps.

### DIFF
--- a/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
+++ b/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
@@ -34,7 +34,30 @@ import sortBy from 'lodash/sortBy';
 import { selectFieldFromApiConfigSchema } from '../../types';
 import { FormHelperText } from '@material-ui/core';
 
-export const SelectFieldFromApi = (props: FieldProps<string>) => {
+const getLabel = (
+  item: unknown,
+  labelSelector: string | string[] | undefined,
+): string | undefined => {
+  if (!labelSelector) {
+    return undefined;
+  }
+  const label = get(item, labelSelector);
+  if (typeof label === 'string') {
+    return label;
+  }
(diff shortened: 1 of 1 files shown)


plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
```diff
@@ -110,12 +132,14 @@ export const SelectFieldFromApi = (props: FieldProps<string>) => {
...
 export const selectFieldFromApiValidation = (
-  value: string,
+  value: unknown,
   validation: FieldValidation,
 ) => {
-  if (!KubernetesValidatorFunctions.isValidObjectName(value)) {
-    validation.addError(
-      'must start and end with an alphanumeric character, and contain only alphanumeric characters, hyphens, underscores, and periods. Maximum length is 63 characters.',
-    );
+  if (typeof value === 'string') {
```
punkle, line 138: Perhaps we should be using the parameter's schema to be validating this field.

plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
```diff
@@ -110,12 +132,14 @@ export const SelectFieldFromApi = (props: FieldProps<string>) => {
...
+    if (!KubernetesValidatorFunctions.isValidObjectName(value)) {
+      validation.addError(
```
punkle, line 140: Honestly.. I think this check is not correct. I dont see why every string returned from the field must be a valid kubernetes object name. It just seems wrong.

I think this validation should look at the schema type and make sure the javascripts types roughly match up.. e.g. a string type returns a string, a number returns a number, a boolean returns a boolean. etc.

sblausten:
Is it possible to test locally?

Issue #980, opened 2023-05-09
ArgoCD returns 500 when retrieving revision metadata


jmadureira:
## Expected Behavior

No error should be present and backstage should display a list of deployments.

## Current Behavior

![Screenshot 2023-05-09 at 08 52 03](https://user-images.githubusercontent.com/770223/237030902-7a5e6d60-896e-4811-9669-873ebd9970ef.png)

## Steps to Reproduce

I'm not entirely certain of all the steps needed to recreate this but this is what I got:
- Backstage component configured with an `argocd/app-selector` annotation
- ArgoCD application installed using an Helm chart pulled from `s3`

## Possible Solution

Not entirely certain since I don't know if this is a incompatibility caused by an argoCD upgrade.

## Context

The 500 error comes when calling the revision metadata endpoint on argoCD so we can easily recreate it using curl:

```bash
curl -v -H "Content-type: application/json" -H "Authorization: Bearer bogus" https://<argoCD instance>/api/v1/applications/<application name>/revisions/<revision id>/metadata
```
... which returns

```
< HTTP/2 500
< date: Tue, 09 May 2023 07:34:19 GMT
< content-type: application/json
< content-length: 98
<
* Connection #0 to host <argoCD instance> left intact
{"error":"revision <revision id> must be resolved","code":2,"message":"revision <revision id> must be resolved"}
```
The `revision id` that is being provided is the Helm chart version (0.0.1 or 0.2.0, etc) of the application that we want to install.

Looking at the [argoCD source code](https://github.com/argoproj/argo-cd/blob/v2.6.7/reposerver/repository/repository.go#LL2187C1-L2187C1) for this endpoint:

```golang
func (s *Service) GetRevisionMetadata(ctx context.Context, q *apiclient.RepoServerRevisionMetadataRequest) (*v1alpha1.RevisionMetadata, error) {
	if !(git.IsCommitSHA(q.Revision) || git.IsTruncatedCommitSHA(q.Revision)) {
		return nil, fmt.Errorf("revision %s must be resolved", q.Revision)
	}
```
it indicates (I think) that this endpoint only supports git commits SHA and not normal version numbers.

Additionally this is a sample of the previous request made to `/api/argocd/argoInstance/<argocd Instance>/applications/selector/<argocd selector annotation>`

```json
{
	"items": [
		{
			"metadata": {
				...
			},
			"spec": {
				...
			},
			"status": {
				"resources": [
					...
				],
				"sync": {
[...]

PR #1039, opened 2023-07-10
fix(argo-frontend-plugin): skip helm revision metadata

janhoon:
In cases where argo application history contains revisions that were deployed via helm, the api calls to `/api/v1/applications/{name}/revisions/{revision}/metadata` would return a 500. This is due to the call requiring the revision param to be a git sha, but in the case of helm it will be a version (and there would be no repo to read from.)

ArgoCD source for the above here: https://github.com/argoproj/argo-cd/blob/master/reposerver/repository/repository.go#L2154-L2156

### DIFF
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
@@ -41,6 +41,9 @@ export const useAppDetails = ({
   ) => {
     const promises: Promise<void>[] | undefined =
       appDetails.status?.history?.map(async (historyRecord: any) => {
+        if (historyRecord.helm !== undefined) {
+          return;
+        }
         const revisionID = historyRecord.revision;
         const revisionDetails = await api.getRevisionDetails({
           url,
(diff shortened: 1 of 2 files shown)


janhoon:
This is a fix for https://github.com/RoadieHQ/roadie-backstage-plugins/issues/980, just need some tests for the changes

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
```diff
@@ -41,6 +41,9 @@ export const useAppDetails = ({
   ) => {
     const promises: Promise<void>[] | undefined =
       appDetails.status?.history?.map(async (historyRecord: any) => {
+        if (historyRecord.source?.chart !== undefined) {
```
janhoon, line 44: From the argo swagger docs for getting an application (`/api/v1/applications/{name}`) for `status.history[].source.helm: 

> Chart is a Helm chart name, and must be specified for applications sourced from a Helm repo.

janhoon:
The result of this PR is ultimately that 2 columns in ArgoCD history is blank if the source is helm:

image

Could add a unit test for this, but don't feel it would be right to check if fields are empty if the source is helm, happy to do it, but will wait for a review

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
janhoon: if this is defined revision will be a helm chart version and not git sha

Xantier (approved):

Issue #1034, opened 2023-07-03
Cannot read properties of undefined (reading 'status') - ArgoCD Plugin Error


dfl-aeb:
"ArgoCD" page throws an error after installation

## Expected Behavior

The argo cd page (or a page where the argo plugin is included) opens as intended

## Current Behavior

```
TypeError
Cannot read properties of undefined (reading 'status')
Call Stack
 eval
  plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDHistoryCard.tsx:57:24
 HistoryTable
  plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDHistoryCard.tsx:56:45
 renderWithHooks
  node_modules/react-dom/cjs/react-dom.development.js:14985:18
 mountIndeterminateComponent
  node_modules/react-dom/cjs/react-dom.development.js:17811:13
 beginWork
  node_modules/react-dom/cjs/react-dom.development.js:19049:16
 HTMLUnknownElement.callCallback
  node_modules/react-dom/cjs/react-dom.development.js:3945:14
 Object.invokeGuardedCallbackDev
  node_modules/react-dom/cjs/react-dom.development.js:3994:16
 invokeGuardedCallback
  node_modules/react-dom/cjs/react-dom.development.js:4056:31
 beginWork$1
  node_modules/react-dom/cjs/react-dom.development.js:23959:7
```
<

## Steps to Reproduce

1. Follow the Plugin installation instructions
2. Configure ArgoCD Api in `app-config.yaml`
4. Navigate to a component's "ArgoCD" page

## Possible Solution

I assume it has something to do with the latest ArgoCD Api version? It stopped working  a few weeks ago.

## Your Environment
- ArgoCD Server v2.7.5+a2430af.dirty
- `@backstage/plugin-catalog-react` version: 1.7.0
- `@backstage/integration` version: 1.5.0
- Tested with Backstage 1.13, 1.14.1 and 1.15.0

## Stacktrace in Developer Console:
<details><summary>Stacktrace in Developer Console:</summary>

The above error occurred in the <HistoryTable> component:

    at HistoryTable (webpack-internal:///../../plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDHistoryCard.tsx:50:25)
    at ArgoCDHistory (webpack-internal:///../../plugins/frontend/backstage-plugin-argo-cd/src/components/ArgoCDHistoryCard.tsx:214:26)
[...]

PR #1042, opened 2023-07-11
null values getting injected for apps with no revision history

rameshbeee:
when using  "argocd/app-selector" the logic to pull the revision history is injecting null values if there is no history available for the argocd app and getting errors in the UI saying cannot read property of undefined, so fixed the issue by filtering out null values from the array
Added undefined vaules in the result list
App with no history in items list
Error from UI
UI after the fix

### DIFF
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
@@ -161,7 +161,9 @@ export const useAppDetails = ({
               return newItem;
             },
           );
-          return await Promise.all(getRevisionHistroyPromises);
+          return Promise.all(getRevisionHistroyPromises)
+            .then(result => result.filter(n => n))
+          ;
         }
         return result.filter(n => n);
       }
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
@@ -144,7 +144,7 @@ export const useAppDetails = ({
           },
         );
         const result = await Promise.all(getRevisionHistroyPromises);
-        return result;
+        return result.filter(n => n);
       }
       if (appSelector || projectName) {
         const result = await api.listApps({ url, appSelector, projectName });
@@ -163,7 +163,7 @@ export const useAppDetails = ({
           );
           return await Promise.all(getRevisionHistroyPromises);
         }
-        return result;
+        return result.filter(n => n);
       }
       return Promise.reject('Neither appName nor appSelector provided');
     } catch (e: any) {
--- a/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
@@ -143,8 +143,8 @@ export const useAppDetails = ({
             return newItem;
           },
         );
-        const result = await Promise.all(getRevisionHistroyPromises);
-        return result.filter(n => n);
+        return Promise.all(getRevisionHistroyPromises)
+          .then(result => result.filter(n => n));
       }
       if (appSelector || projectName) {
         const result = await api.listApps({ url, appSelector, projectName });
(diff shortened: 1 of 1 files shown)


plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
```diff
@@ -140,7 +140,7 @@ export const useAppDetails = ({
           },
         );
         const result = await Promise.all(getRevisionHistroyPromises);
-        return result;
+        return result.filter(n => n);
```
planeiii-te, line 143: This is already a promise so there is need to `await` this. This is effectively `return await` which is redundant. 

```suggestion
        return Promise.all(getRevisionHistroyPromises)
          .then(result => result.filter(n => n))
        ;
```

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
```diff
@@ -159,7 +159,7 @@ export const useAppDetails = ({
           );
           return await Promise.all(getRevisionHistroyPromises);
```
planeiii-te, line 160: `return await` is a redundant. The caller already has to await this.

```suggestion
          return Promise.all(getRevisionHistroyPromises)
            .then(result => result.filter(n => n))
          ;
```

planeiii-te (changes requested):
A few clean up suggestions, but I'd be happy to see this fixed soon. I had a tough time tracking this down.

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
rameshbeee: Thanks for your suggestions, couldn't commit from here, so pushed from my branch

planeiii-te (approved):
LGTM

jpilowsk-te:
This would probably resolve https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1034

planeiii-te (approved):
This looks good to me, but I don't maintain this project.

bbecker-te (approved):

planeiii-te:
@rameshbeee Can you fix the `prettier` and `tsc` issues in the failing checks so this will be ready to merge whenever they finally see it?

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
```diff
@@ -139,8 +139,8 @@ export const useAppDetails = ({
             return newItem;
           },
         );
-        const result = await Promise.all(getRevisionHistroyPromises);
-        return result;
+        return Promise.all(getRevisionHistroyPromises)
```
Xantier, line 142: Let's not move away from using `await` which should be the default approach to handling Promises.

Xantier (changes requested):
This is missing a changeset

plugins/frontend/backstage-plugin-argo-cd/src/components/useAppDetails.ts
planeiii-te: @Xantier `return await` is an anti-pattern outside of `try-catch`. The [ESLint rule](https://eslint.org/docs/latest/rules/no-return-await) that prevents it explains why. Also here's a [blog post](https://jakearchibald.com/2017/await-vs-return-vs-return-await/) that goes a bit more in depth

Xantier: Thanks for the blog post. An alternative approach to `then` should be used regardless.

rameshbeee:
> @rameshbeee Can you fix the `prettier` and `tsc` issues in the failing checks so this will be ready to merge whenever they finally see it?

@planeiii-te, could you help me with the prettier and tsc issues? I am new to ts

planeiii-te:
> could you help me with the prettier and tsc issues? I am new to ts

@rameshbeee Your force pushing is erasing the test runs so I can no longer see what was failing. Previously you could see the checks failing and they would tell you what the problem was with the code. It's hard to diagnose without those errors. 

Also as the maintainer has stated:

> This is missing a changeset

Just run `yarn changeset` and it should walk you through the process and generate a file that you will have to commit in the `.changeset` folder.

PR #1043, opened 2023-07-12
Display confluence changes in activity stream

stephane-mori:
Hello,

I can observe that plugin jira don't allow confluence changes to be displayed with activity stream : 
https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/frontend/backstage-plugin-jira/src/api/index.ts#L333

Display this changes could be interresting for me. That's why I want to provide a patch allowing to configure the jiraCard component to display confluence changes with activity stream.

Regards,
Stéphane MORI

### DIFF
--- a/plugins/frontend/backstage-plugin-jira/src/components/JiraCard/JiraCard.tsx
+++ b/plugins/frontend/backstage-plugin-jira/src/components/JiraCard/JiraCard.tsx
@@ -76,10 +76,11 @@ const CardProjectDetails = ({
 
 type JiraCardOptionalProps = {
   hideIssueFilter?: boolean;
+  displayConfluenceChanges?: boolean;
 };
 
 export const JiraCard = (props: EntityProps & JiraCardOptionalProps) => {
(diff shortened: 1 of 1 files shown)


sblausten (changes requested):
Hi, I'm not sure I understand this change you've made - it seems to just give the option to leave out the tickets param - or rather it makes that the default which seems like a major breaking change.

If you wanted to change the activityStream filter that you linked to you would probably want to pass in some kind of flag here https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/frontend/backstage-plugin-jira/src/components/JiraCard/components/ActivityStream.tsx#L137 to the hook and then add some logic to reduce the filtering if the flag is true. Does that make sense?

punkle:
@stephane-mori I think it might be a little more clear if you were to add a new option to the getActivityStream function. It could be called "filterConfluenceActivity" and default it to true.. What do you think?

PR #1049, opened 2023-07-17
Allow more RSS Feed customization

ivangonzalezacuna:
In this PR I tried to accomplish a bunch of things.

1. First, I added a `title` to the homeCard definition, otherwise the widget name is empty in the new custom homepage, as well as updated a deprecated reference
2. Then, I tried to add some more customization. For our use case for instance, we would like to add the author to each row. So I went for the "easiest" solution, which is allowing to customize the whole rendering. If not defined, then it will stay as before and nothing will break. And it's promise, since this allows you to make any other API calls if needed.
3. I also removed the `title` and `Box` for the table. Since the new custom homepage is already adding a header, it seems a bit too repetitive. 

This is not going to break anything, but just add a bit more of functionality to the component, as well as cleaning up a little bit the previous code.

If you have any concerns, just let me know.


.changeset/famous-teachers-breathe.md
```diff
@@ -0,0 +1,9 @@
+---
+'@roadiehq/backstage-plugin-home-rss': patch
```
Xantier, line 2: It introduces some new functionality so we can happily make this a minor bump.

plugins/home/backstage-plugin-home-rss/package.json
```diff
@@ -35,7 +35,7 @@
     "@backstage/config": "^1.0.8",
     "@backstage/core-components": "^0.13.2",
     "@backstage/core-plugin-api": "^1.5.2",
-    "@backstage/plugin-home": "^0.5.3",
```
Xantier, line 38: This modification should contain the related changes to a yarn.lock file.

Xantier (commented):
Few minor tweaks and this should be good to go

plugins/home/backstage-plugin-home-rss/src/RSSCard/types.ts
```diff
@@ -14,11 +14,24 @@
...
+};
+
 /**
  * Props for RSS content component {@link Content}.
  *
  * @public
  */
 export type RSSContentProps = {
   feedURL: string;
+  paging?: boolean;
```
punkle, line 33: Would it make sense to add these to the readme file?

ivangonzalezacuna: Indeed it does. I added a few more lines to the README file

plugins/home/backstage-plugin-home-rss/package.json
ivangonzalezacuna: I don't know if I did something wrong. But I executed `yarn install` and the yarn.lock didn't change at all. I even tried removing the node_modules inside this plugin

.changeset/famous-teachers-breathe.md
ivangonzalezacuna: Done!

plugins/home/backstage-plugin-home-rss/src/RSSCard/Content.tsx
```diff
@@ -54,13 +50,7 @@ const skeletonDataItem = {
...
-  skeletonDataItem,
-  skeletonDataItem,
-  skeletonDataItem,
-  skeletonDataItem,
-  skeletonDataItem,
-];
+const skeletonData = Array(6).fill(skeletonDataItem);
 
 /**
  * A component to render a RSS feed
```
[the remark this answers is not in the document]
ivangonzalezacuna: I applied this change as well

plugins/home/backstage-plugin-home-rss/package.json
Xantier: That's no worries. It could be that this dep is already included in the with the same versioning

Xantier (approved):
Update the branch against main just in case and after that we can ship this.

ivangonzalezacuna:
> Update the branch against main just in case and after that we can ship this.

The branch should be up-to-date with main.

PR #1053, opened 2023-07-18
correct test descriptions

mbenson:
#### :heavy_check_mark: Checklist

- [ ] Added tests for new functionality and regression tests for bug fixes
- [X] Added changeset (run `yarn changeset` in the root)
- [ ] Screenshots of before and after attached (for UI changes)
- [ ] Added or updated documentation (if applicable)

PR #1054, opened 2023-07-18
Object results

mbenson:
This PR builds on https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1053 which I hope will be accepted. I noticed that the result of the jsonata actions was documented as being of `object` type even though the Typescript seemed pretty adamant that the type was `string`. This PR adds an additional `as` input parameter which, when supplied the value `object` will set the action's result as the actual object returned from executing the JSONata `expression` against the specified `data` parameter, rather than encoding it back to the original format. Omitting `as` or specifying  `string` will trigger the original functionality.

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/json.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/json.ts
@@ -50,6 +50,7 @@ export function createJsonJSONataTransformAction() {
             description:
               'Permitted values are: "string" (default) and "object"',
             type: 'string',
+            enum: ['string', 'object'],
           },
           replacer: {
             title: 'Replacer',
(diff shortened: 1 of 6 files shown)


mbenson:
Note that the effect of this PR could probably be achieved using multiple existing roadie utils actions (`fs:parse` + `jsonata`) in which case an alternative would be to correct the declared output type of the existing implementations to `string`.

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/json.ts
```diff
@@ -44,6 +45,12 @@ export function createJsonJSONataTransformAction() {
             description: 'JSONata expression to perform on the input',
             type: 'string',
           },
+          as: {
+            title: 'Desired Result Type',
+            description:
+              'Permitted values are: "string" (default) and "object"',
+            type: 'string',
```
Xantier, line 52: We could possibly add `enum: [the types]` here also?

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/json.ts
```diff
@@ -64,26 +71,29 @@ export function createJsonJSONataTransformAction() {
...
           result: {
             title: 'Output result from JSONata',
-            type: 'object',
+            type: 'object | string',
           },
         },
       },
     },
     async handler(ctx) {
+      let post: (rz: any) => any;
```
Xantier, line 80: maybe rename this to be a little bit more descriptive

mbenson: Later I saw how the file formats on `fs:parse` are handled with a `Record` but I don't know that that would be particularly helpful in this scenario. I will expand on the variable name.

Xantier (approved):

PR #1076, opened 2023-08-08
Bump up Argo CD plugin version

jpilowsk-te:
Bump up version of Argo CD Plugin so that changes in PR #1042  become available


kissmikijr:
We use changeset to bump the package version.

PR #1088, opened 2023-08-24
Passthrough param in http request from previous input

sblausten:
Adds the ability to toggle between and parse different http requests depending on previously selected parameters.

https://www.loom.com/share/f87bf04679c342ffb32a72f52b9e3dc3

https://app.shortcut.com/larder/story/16086/create-dependant-dynamic-parameter-steps-in-scaffolder

e.g.:

```
---
apiVersion: scaffolder.backstage.io/v1beta3
kind: Template
metadata:
  name: http-param-new
  title: Http param test
spec:
  owner: roadie
  type: service
  parameters:
    properties:
      objectFacet:
        title: 'Facet Mapping'
        enumNames:
          - "Kind"
          - "Name"
        enum:
          - paramValue: 'kind'
            paramKey: 'facet'
            path: 'catalog/entity-facets'
            jsonata: 'facets.kind.value'
          - paramValue: "group"
            paramKey: "filter=kind"
            path: 'catalog/entities'
            jsonata: 'metadata.name'
      entities:
        type: string
        ui:field: SelectFieldFromApi
        ui:options:
          title: Dependant select
          dynamicParams:
            paramKeyLocation: "objectFacet.paramKey"
            paramValueLocation: "objectFacet.paramValue"
            pathLocation: "objectFacet.path"
            jsonataLocation: 'objectFacet.jsonata'
```

### DIFF
--- a/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
+++ b/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
@@ -187,8 +187,10 @@ properties:
       - 'Entity name'
     enum:
       - facet: 'kind'
+        paramKey: 'facet'
         jsonata: 'facets.kind.value'
       - facet: 'metadata.name'
+        paramKey: 'facet'
         jsonata: 'facets.`metadata.name`.value'
   custom:
     type: string
@@ -197,11 +199,11 @@ properties:
       title: Select Populated Depending on Previous field
       # The Path to the Backstage API
       path: 'catalog/entity-facets'
-      # (Optional) Specifies the key for the query param that appends the path i.e. `/catalog/entity-facets?facet=kind
-      params:
-        facet: 'kind'
       # (Optional)
       dynamicParams:
+        # (Optional) The key of a previous template property who's selected value will be used as an additional query parameter value on the request
+        # i.e. '?facet=' or '?filter=kind=' depending on whether `Kind` or `Group name` is selected.
+        paramKeyLocation: 'objectFacet.paramKey'
         # (Optional) The key of a previous template property in this form who's selected value will be used as the query parameter value on the request
         paramValueLocation: 'facetMapping'
         # (Optional) The path to a previous template property for the jsonata string that will be used to parse the request's body and extract values.


plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
```diff
@@ -79,3 +79,123 @@ The example template would result in the following dropdown:
...
+          # This selects the array element from the API fetch response. It finds the array with the name kind
+          # under the facets object
+          arraySelector: 'facets.kind'
+          # (Optional) This selects the field in the array to use for the value of each select item. If its not specified
+          # it will use the value of the item directly.
+          valueSelector: 'count'
+          # (Optional) This selects the field in the array to use for the label of each select item.
+          labelSelector: 'value'
+          # (Optional)
+          dynamicParams:
```
punkle, line 124: I wonder is it possible to use the existing params and allow substitution like this. If we wanted to go with this option maybe the `$` is a bad choice.. because it might be confused with the substitution that exists in the template steps.

```
params:
  facet: '${{parameters.facetValue}}'
  filter:
    kind: 'Component'
```

plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
```diff
@@ -79,3 +79,123 @@ The example template would result in the following dropdown:
...
+            # (Optional) This is the key of an additional dynamic query parameter that can be added to the request
+            paramKeyLocation: 'facet'
+            # (Optional) This is the key of a previous property who's selected value will be used as an additional query parameter value on the request
```
punkle, line 127: This sentence is a little hard to get my head around. Maybe consider splitting it in two?

plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
punkle: I suppose you could also use the same substitution for path, value selector, label selector, array selector etc

sblausten: I tried that but it doesn't seem to work

punkle: Its probably not as simple as this:

```javascript
const options = {
  // this mustache template would be passed in via the scaffolder template options
  path: "proxy/my-api/{{params.path}}"
}


const render(input: unknown, data) {
  if (typeof input === 'object') {
    return Object.keys(input).map(key => [key, renderRecursive(input[key])]);
  }
  if (typeof input === 'string') {
    return Mustache.render(input, data);
  }
  return input
}

const path = render(options.path, { paramaters: formData });
const params = renderRecursive(options.params, { paramaters: formData });
// and so on.
```

plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
```diff
@@ -79,3 +79,135 @@ The example template would result in the following dropdown:
...
```
punkle, line 149: This sentence is very complicated. I am not sure I actually understand what it means. Maybe the previous comment could do with simplying also. Maybe a suggestion: in the paragraph above this example, explain the scenario with concrete examples of data and then use them in this code example.

plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
sblausten: Sorry but I don't understand what this is - where would this go?

punkle: In the end I would be able to do something like this I mean:

```
  parameters:
    properties:
      - title: Custom
        properties:
          facet:
            type: string
          path:
            type: string
          valueSelector:
            type: string
          labelSelector:
            type: string
          custom:
            type: string
            ui:field: SelectFieldFromApi
            ui:options:
              title: My Dropdown title
              description: My custom description for the component
              path: '{{ parameters.path }}'
              params:
                facet: '{{ parameters.facet }}'
              arraySelector: 'facets.{{ parameters.facet }}'
              valueSelector: '{{ parameters.valueSelector }}'
              labelSelector: '{{ parameters.labelSelector }}'
```

sblausten: Do you want to try out your approach and see if it works? I'm not sure I get where the code you wrote should go

punkle: https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1115

punkle: Here is a demo of that https://www.loom.com/share/6f0dd471728d4cefa23f74509945c2a5

PR #1096, opened 2023-08-29
Improvements to Bugsnag plugin

fhp:
This PR fixes some issues:
* It currently throws an error if the annotation is missing, the fixes that so it shows the 'please add the annotation' message again.
* The stages of an error that bugsnag returns are an array. They are now joined with a comma.

And it improves the display of data in the table:
* It adds the description and the status to the table. Both seem rather important to me.
* It adds filtering options to the table. I had to remove the title here as it get rendered really weirdly otherwise (aligned to the right).

I have one additional change that I'd like to make, but that is a more intrusive change, so I'd like to get some feedback on that: The plugin is currently configured with the `bugsnag.com/project-key` and `bugsnag.com/project-name` annotation, where the project-key annotation contains something like `Organization name/projectApiKey`. As the api-key is a secret, I'd rather not have it my repository.

This change allows (optionally) changing the content of the `bugsnag.com/project-key` annotation to `Organization name/Project name`. As this already contains the project name, the `bugsnag.com/project-name` annotation becomes deprecated. Configuring the plugin this way makes more sense, as the secret api-key can be kept out of the repositories.

The plugin will detect if the plugin is configured with an old annotation, and revert to the old behaviour if needed.

I would have preferred to use the slugs, but it is not possible to use the slug in the search `q` parameter in the [Bugsnag API](https://bugsnagapiv2.docs.apiary.io/#reference/projects/projects/list-an-organization's-projects).

- [x] Screenshots of before and after attached (for UI changes) -> in files changed


.changeset/eighty-mice-know.md
Xantier, line 1: We can merge these changesets into a single minor one I think

Xantier:
> I have one additional change that I'd like to make, but that is a more intrusive change, so I'd like to get some feedback on that: The plugin is currently configured with the bugsnag.com/project-key and bugsnag.com/project-name annotation, where the project-key annotation contains something like Organization name/projectApiKey. As the api-key is a secret, I'd rather not have it my repository.

I think this is fine (and welcomed!) change. Let's keep backwards compatibility for now with the old annotation structure but the way you are hiding it in the documentation is probably a good incentive to push people to migrate

.changeset/eighty-mice-know.md
fhp: Alright, I'll merge them.

Xantier (approved):

Issue #900, opened 2023-03-21
`catalog-backend-module-okta` User/Group Entity Providers allow custom Entity Descriptor Callbacks


alecjacobs5401:
It would be great to have full control over the emitted `UserEntity` and `GroupEntity` entities that are generated from the `OktaUserEntityProvider` and `OktaGroupEntityProvider` entity providers respectively.

This could be a fully backwards compatible and default to the currently generated `UserEntity` and `GroupEntity` definitions.

## Feature Suggestion

* `OktaUserEntityProvider` should provide a new optional argument on instantiation `entityFromOktaUser`
* `OktaGroupEntityProvider` should provide a new optional argument on instantiation `entityFromOktaGroup`

Doing so will allow for customization of emitted Entities for Backstage instances that have extensions to the default Catalog schema (e.g. custom relations). Additionally, it allows for just greater configuration of Entities based on the Okta Resources.

## Possible Implementation

For `OktaUserEntityProvider`, for example:

```diff
+import type { User as OktaUser } from '@okta/okta-sdk-nodejs';
+export type OktaUserEntityGenerator = (user: OktaUser, options: { annotations: Record<string, string> }) => Entity   // specifically Entity for custom extensions

export class OktaUserEntityProvider extends OktaEntityProvider {
  private readonly namingStrategy: UserNamingStrategy;
+ private readonly entityFromOktaUser: OktaUserEntityGenerator;
  private readonly userFilter?: string;
  private readonly orgUrl: string;

  static fromConfig(
    config: Config,
    options: {
      logger: winston.Logger;
      namingStrategy?: UserNamingStrategies | UserNamingStrategy;
+     entityFromOktaUser?: OktaUserEntityGenerator;
    },
  ) {
    const accountConfig = getAccountConfig(config);

    return new OktaUserEntityProvider(accountConfig, options);
  }

  constructor(
    accountConfig: AccountConfig,
    options: {
      logger: winston.Logger;
      namingStrategy?: UserNamingStrategies | UserNamingStrategy;
+     entityFromOktaUser?: OktaUserEntityGenerator;
    },
  ) {
    super([accountConfig], options);
    this.namingStrategy = userNamingStrategyFactory(options.namingStrategy);
    this.userFilter = accountConfig.userFilter;
    this.orgUrl = accountConfig.orgUrl;
+   this.entityFromOktaUser = options.entityFromOktaUser
  }

  async run(): Promise<void> {
    if (!this.connection) {
[...]

PR #1103, opened 2023-09-04
Add option to provide custom user transformer

ianaktr:
This PR adds an option to provide a custom user transformer for okta users in user and org providers. It is a similar feature that LDAP has. If the transformer is provided it will be used instead of [userEntityFromOktaUser](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/backend/catalog-backend-module-okta/src/providers/userEntityFromOktaUser.ts).

In our use case, we want to fill some of the fields of the user entity which isn't possible to do now. For example,  `profile.picture` or `metadata.description`.

Something similar was discussed in this issue https://github.com/RoadieHQ/roadie-backstage-plugins/issues/900

If this approach is accepted, I can also do that for groups.


kissmikijr:
Fantastic thank you for this. It looks great. Would you mind adding to the groups as well?

ianaktr:
Yes, of course! I will add this for groups as well 👍

ianaktr:
@kissmikijr Hey! I added functionality to the groups as well and also updated the documentation. Would be nice if you could take a look once you have time. Thanks!

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
```diff
@@ -422,5 +424,56 @@ describe('OktaGroupProvider', () => {
...
+      await provider.run();
+      expect(entityProviderConnection.applyMutation).toBeCalledWith({
+        type: 'full',
+        entities: expect.arrayContaining([
+          expect.objectContaining({
+            entity: expect.objectContaining({
+              kind: 'Group',
+              metadata: expect.objectContaining({
+                name: 'asdfwefwefwef',
+                description: 'Everyone in the company',
```
kissmikijr, line 465: Shouldn't we add the annotations to this as well?

plugins/backend/catalog-backend-module-okta/src/providers/types.ts
```diff
@@ -0,0 +1,39 @@
...
+  Group as OktaGroup,
+  User as OktaUser,
+} from '@okta/okta-sdk-nodejs';
+import { GroupNamingStrategy } from './groupNamingStrategies';
+import { UserNamingStrategy } from './userNamingStrategies';
+
+export type OktaGroupEntityTransformer = (
+  group: OktaGroup,
+  namingStrategy: GroupNamingStrategy,
+  parentGroup: OktaGroup | undefined,
```
kissmikijr, line 28: I think we could make parentGroup optional so we dont need to write out the explicit undefined here
```suggestion
  parentGroup?: OktaGroup,
```

kissmikijr (commented):
Thank you! Left some tiny tiny comments, otherwise it looks good!

plugins/backend/catalog-backend-module-okta/src/providers/types.ts
ianaktr: Good catch! Done

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.test.ts
ianaktr: Sorry, I am not sure I understand. Do you mean to check annotations object in metadata? The untransformed group doesn't have any custom annotations that's why I didn't add any checks here. Only for the new `profile.displayName` which isn't available in the default transformer

kissmikijr (approved):
Thank you! Let's ship it!

PR #1107, opened 2023-09-08
Allow for multiple dashboard urls to be provided and rendered for a c…

jsidney:
The current Datadog plugin only allows for a single dashboard to be rendered through the `datadoghq.com/dashboard-url:` annotation.  There are going to be use-cases where a certain component will have multiple dashboards that need to be displayed through the plugin. This change allows for a comma-separated list of dashboard URLs to be passed through this annotation and rendered on the component page.

### DIFF
--- a/plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
+++ b/plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
@@ -17,33 +17,41 @@
 import React from 'react';
 import { Entity } from '@backstage/catalog-model';
 import { InfoCard } from '@backstage/core-components';
+import { Grid } from '@material-ui/core';
 import { useDatadogAppData } from './useDatadogAppData';
 import { Resizable } from 're-resizable';
 import ZoomOutMapIcon from '@material-ui/icons/ZoomOutMap';
 
 export const DatadogDashboardPage = ({ entity }: { entity: Entity }) => {
   const { dashboardUrl } = useDatadogAppData({ entity });
+  const allDashboardUrls: string[] = dashboardUrl.split(',');
   return (
-    <InfoCard title="Datadog dashboard">
-      <Resizable
-        defaultSize={{
-          width: '100%',
-          height: 500,
-        }}
-        handleComponent={{ bottomRight: <ZoomOutMapIcon /> }}
-      >
-        <iframe
-          title="dashboard"
-          src={`${dashboardUrl}`}
-          style={{
-            border: 'none',
-            height: '100%',
-            width: '100%',
-            resize: 'both',
-            overflow: 'auto',
-          }}
-        />
-      </Resizable>
-    </InfoCard>
+    <Grid container spacing={3}>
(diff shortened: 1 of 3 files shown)


jsidney:
## Current display: 
datadog_current_screenshot

## Proposed change:
datadog_proposed_screenshot

plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
```diff
@@ -17,33 +17,41 @@
...
-            width: '100%',
-            resize: 'both',
-            overflow: 'auto',
-          }}
-        />
-      </Resizable>
-    </InfoCard>
+    <Grid container spacing={3}>
+      {allDashboardUrls.map((value, index) => (
+        <Grid item md={12}>
```
Xantier, line 31: Could you add a key to this component?

jsidney: @Xantier  - i assume by key, you mean like and index? (eg: name of the dashboard or something?)

jsidney:
@Xantier  - i want to add to the mock so that the tests for this plugin are also up to date.... my question is that I cant see how the `plugin.test.tsx` calls the instantiates the mock. so before I add to the mock, I want to just make sure I do it correctly. can you please point me in the right direction so that I can add the tests correctly?

plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
Xantier: More info: https://react.dev/learn/rendering-lists. the name or maybe the annotation itself should work here.

Xantier:
@jsidney, it looks like Datadog plugin doesn't have tests written for it yet. You can probably take a look at an example from argocd plugin for example and see how they handle the mocks in that package.

plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
jsidney: i have added the key 👍

Xantier (approved):

plugins/frontend/backstage-plugin-datadog/src/components/DatadogDashboardPage.tsx
```diff
@@ -17,33 +17,41 @@
...
+          <InfoCard title={`Datadog dashboard ${index}`} variant="gridItem">
+            <Resizable
+              defaultSize={{
+                width: 100,
```
stasOR41k, line 35: should it be reverted to the '100%'?
I have an issue with the default size in the [2.1.15](https://www.npmjs.com/package/@roadiehq/backstage-plugin-datadog/v/2.1.15)

jsidney: More than happy to make that change @stasOR41k  - I think it must have been something I had changed without noticing while playing around with the look and feel

PR #1135, opened 2023-09-27
Added package installation instructions

stephanschielke:
Also switched example from string selector to array selector since `facets.spec.type` for example would not work and result in `[facets, spec, type]` rather than `[facets, spec.type]`

### DIFF
--- a/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
+++ b/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
@@ -54,9 +54,7 @@ spec:
 
               # This selects the array element from the API fetch response. It finds the array with the name kind
               # under the facets object
-              arraySelector:
-                - facets
-                - kind
+              arraySelector: 'facets.kind'
 
               # (Optional) This selects the field in the array to use for the value of each select item. If its not specified
               # it will use the value of the item directly.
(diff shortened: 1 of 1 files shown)


plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/README.md
```diff
@@ -49,7 +54,9 @@ spec:
 
               # This selects the array element from the API fetch response. It finds the array with the name kind
               # under the facets object
-              arraySelector: 'facets.kind'
+              arraySelector:
```
punkle, line 57: This change is incorrect. It should be `arraySelector: 'facets.kind'`

stephanschielke: The problem I was facing is that:
```
          arraySelector: 'facets.spec.type'
```
for example, gets interpreted as:
```
          arraySelector:
            - facets
            - spec
            - type
```
rather than
```
          arraySelector:
            - facets
            - spec.type
```        

If you explicitly configure an array as the selector rather than using the string selector you avoid having this problem. Type-wise both are allowed:
https://github.com/RoadieHQ/roadie-backstage-plugins/blob/d5b537e97985a1bc3f1d8d7997d1765e0eb68736/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/types.ts#L21

PR #1160, opened 2023-10-23
Return entities after validating

mrooding:
I have a use case where I'd like to use your validator but I'd also like to do additional checks based on the parsed entities. I could copy paste a large part of your util to parse them once again but simply returning the objects also works for me.


utils/roadie-backstage-entity-validator/src/validator.test.js
kissmikijr, line 1: Could you make these tests to compare it to the expected value instead of the length of the returned array ?

mrooding: @kissmikijr I've updated the PR

Xantier (approved):

PR #1164, opened 2023-10-25
Allow SelectFieldFromApi to use OAuth

iain-b:
Add an optional prop to the component which can be used to enable OAuth login. For example, if a user supplies oauth={ provider: 'github', scopes:  ['repo'] } then we get an access token for GitHub using the backstage GitHub auth API and pass that through via the Authorization header. This requires/ assumes a proxy entry has been configured to pass through the header.


Xantier (dismissed):
Looks to be a (relevant) failing test. The logic and code looks fine

iain-b:
> Looks to be a (relevant) failing test. The logic and code looks fine

TY Had to pull in recent changes which added a new prop and needed to add that for the test also.

Xantier (approved):

PR #1187, opened 2023-11-17
Return error responses as well as successful responses in templates

sblausten:
https://app.shortcut.com/larder/story/17453/einride-fetch-entities-from-scaffolder-template

This will allow logic that can expect 404s from an endpoint and continue the scaffolder job rather than erroring out.

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
@@ -34,6 +34,7 @@ export function createHttpBackstageAction(options: {
     params?: Params;
     body?: any;
     logRequestPath?: boolean;
+    continueOnBadResponse?: boolean;
   }>({
     id: 'http:backstage:request',
     description:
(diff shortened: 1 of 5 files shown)


plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/helpers.ts
```diff
@@ -78,7 +78,7 @@ export const http = async (
         res.status
       } Response body: ${JSON.stringify(body)}`,
     );
-    throw new HttpError('Unable to complete request');
+    return { code: res.status, headers: {}, body };
```
sblausten, line 81: Blanking the headers for security reasons

punkle, line 81: I wonder would any code already depend on the current behaviour?

Maybe it makes sense to add a new option to the action step like the axios `validateStatus: false`.

Otherwise, we should consider releasing this as a major update.

I prefer the validateStatus option.

sblausten: Yes good ideas! I'll add the option

plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
```diff
@@ -85,6 +86,13 @@ export function createHttpBackstageAction(options: {
               'Option to turn request path logging off. On by default',
             type: 'boolean',
           },
+          continueOnBadResponse: {
+            title: 'Continue on error',
+            description:
+              'Return response code and body and continue to next scaffolder step if the response status is not a 2xx. False by default.',
```
punkle, line 92: 301 is also a successful response. Can I suggest the following wording.

"Return response code and body and continue to next scaffolder step if the response status shows an error. By default the step will fail if any status code is returned 400 and above."

punkle (approved):

PR #1198, opened 2023-12-01
Use the CORE_SCHEMA when parsing entity YAML

brianphillips:
By default (an despite the [recommendations in the YAML spec](https://yaml.org/spec/1.2-old/spec.html#id2804923)), `js-yaml` [uses](https://www.npmjs.com/package/js-yaml#load-string---options-) a `DEFAULT_SCHEMA` which supports additional "special" types in the YAML document. This behavior differs from the `yaml` package (used by the core Backstage catalog processor/parser) which [defaults](https://eemeli.org/yaml/#schema-options) to the `CORE_SCHEMA` instead.

This change adjusts the usage of `js-yaml` to more closely reflect the actual YAML processor used by Backstage so that (for instance) if an annotation has a value that looks like a date, the value is parsed as a string (like `yaml` does) instead of a `Date` object (which is the default behavior under the `DEFAULT_SCHEMA` when parsing with `js-yaml`).

### DIFF
--- a/.changeset/nine-olives-deny.md
+++ b/.changeset/nine-olives-deny.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/roadie-backstage-entity-validator': patch
+---
+
+use CORE_SCHEMA when processing entity YAML to more closely reflect how Backstage core parses YAML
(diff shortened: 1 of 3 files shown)


brianphillips:
Could I get a review?

Xantier (approved):

PR #1201, opened 2023-12-06
Add option to merge array values 

iain-b:
Add option to merge array values in merge actions. This means rather than being overwritten values in the new array value will be concatenated onto the target array. This behvaiour is optional and is disabled by default to maintain backward compatibility. 


plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/merge/merge.ts
```diff
@@ -18,12 +18,23 @@ import { createTemplateAction } from '@backstage/plugin-scaffolder-node';
...
-import { merge } from 'lodash';
+import { isArray, isNull, mergeWith } from 'lodash';
 import yaml from 'js-yaml';
 import { supportedDumpOptions, yamlOptionsSchema } from '../../types';
 
+function mergeArrayCustomiser(objValue: string | any[], srcValue: any) {
+  if (isArray(objValue) && !isNull(objValue)) {
+    return Array.from(new Set(objValue.concat(srcValue)));
+  }
+  return undefined;
```
Xantier, line 29: would it be better to return the original value in here instead of undefined?

Issue #1209, opened 2023-12-29
roadiehq:utils:json:merge allow for custom spacing


dlaird-ovo:
Currently this scaffolder action is hardcoded to a 2 space indent. 
We are modifying existing JSON files where the indent is specified as 4.

If I submitted a PR with an option the change the default spacing from 2, to a custom value would that be ok?

PR #1210, opened 2023-12-29
Allow for the JSON file output after a 'merge` to match input file indentation.

dlaird-ovo:
This supports a new optional parameter with the `roadiehq:utils:json:merge` action:
* `matchFileIndent` 

This will check the source file and identify the indentation used - and then if set to true will write the output file with the same indentation. 
Useful where modifying existing JSON files with an indentation other than '2'

The default and fallback value it still '2'

This is a PR related to this issue: #1209 

I have added a simple additional test where it has a source file with 4 space indent, and checks the output file is also 4.
It checks that when the parameter is not set the output file indentation remains at '2'.

### DIFF
--- a/.changeset/large-roses-dance.md
+++ b/.changeset/large-roses-dance.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': minor
+---
+
+Allow option for JSON output files after merging to match the input file indentation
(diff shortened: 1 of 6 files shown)


sblausten (commented):
Nice! Looks good. 

Could you add the optional param to the README here also https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/scaffolder-actions/scaffolder-backend-module-utils/README.md#merge-json and just explain briefly what it means? Thanks

sblausten (approved):

Issue #1181, opened 2023-11-16
ArgoCD app selector appears to be broken in latest version


dwilliams782:
With the latest ArgoCD plugin (2.4.0), when specifying `argocd/app-selector` annotation, this is ignored and will show all applications after querying every revision even when just using the Overview card (I see a bug for this already: https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1121) . 

I downgraded the plugin to 2.2.10 and it works normally. EDIT: Also tried latest-1, 2.3.6, which is good as well.

## Expected Behavior

With a component like:

```
apiVersion: backstage.io/v1alpha1
kind: Component
metadata:
  annotations:
    argocd/app-selector: "app=sysops-db-tools"
```

The filter should be passed to the argocd proxy i.e.

```
[1] 2023-11-16T13:08:51.841Z backstage info ::1 - - [16/Nov/2023:13:08:51 +0000] "GET /api/proxy/argocd/api/applications?selector=app%3Dsysops-db-tools HTTP/1.1" 200 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36" type=incomingRequest
```

Above log taken from working 2.2.10 plugin.

## Current Behavior

The selector is ignored and we see all applications in the ArgoCD Overview.

```
[1] 2023-11-16T13:16:31.145Z backstage info ::1 - - [16/Nov/2023:13:16:31 +0000] "GET /api/proxy/argocd/api/applications HTTP/1.1" 200 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36" type=incomingRequest

```

![image](https://github.com/RoadieHQ/roadie-backstage-plugins/assets/42545485/18144b7e-e77c-4ace-923a-772abbd7f304)

Above logs taken with 2.4.0 installed.

## Steps to Reproduce

Install ArgoCD plugin 2.4.0 using the instructions found [here](https://github.com/RoadieHQ/roadie-backstage-plugins/tree/main/plugins/frontend/backstage-plugin-argo-cd).

## Your Environment

- `@backstage/plugin-catalog-react` version: ^1.8.5
- `@backstage/integration` version: ^1.7.1

dwilliams782:
[...]

PR #1126, opened 2023-09-21
Adds support for Applications in any namespace

mwienk:
I've added support for namespaced applications in Argo CD (issue #1068 and #1038). I've added it all under a feature flag argocd.namespacedApps.

I do think it needs documentation for the functionality. At least the part enabling the functionality:
```yaml
[app-config.yaml]
argocd:
  namespacedApps: true
```

And how to use new annotation:
```yaml
[entity.yaml]
...
metadata:
  annotations:
    argocd/app-namespace: my-test-ns
```

should be available somwhere, but I'm not sure if you want it in the readme file or perhaps somewhere else to keep things clean.

#### :heavy_check_mark: Checklist

- [X] Added tests for new functionality and regression tests for bug fixes
- [X] Added changeset (run `yarn changeset` in the root)
- [ ] Screenshots of before and after attached (for UI changes)
- [ ] Added or updated documentation (if applicable)

PR #1212, opened 2023-12-30
fix: ensure query params are preserved when not using namespaced apps

alecjacobs5401:
Resolves https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1181

I'm not 100% sure what the intent of with the param filtering is, but based on context provided from https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1126, I _believe_ that the intent is to remove the `appNamespace` query param if the Backstage instance is not opted into namespaced app support via `app-config.yaml`.

- [x] Screenshots of before and after attached (for UI changes) **N/A**
- [x] Added or updated documentation (if applicable) **N/A**

### DIFF
--- a/plugins/frontend/backstage-plugin-argo-cd/src/api/index.ts
+++ b/plugins/frontend/backstage-plugin-argo-cd/src/api/index.ts
@@ -19,7 +19,6 @@ import {
   isDecodeError as tsIsDecodeError,
 } from 'io-ts-promise';
 import reporter from 'io-ts-reporters';
-import { ARGOCD_ANNOTATION_APP_NAMESPACE } from '../components/useArgoCDAppData';
 
 export interface ArgoCDApi {
   listApps(options: {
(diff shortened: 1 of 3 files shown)


plugins/frontend/backstage-plugin-argo-cd/src/api/index.ts
```diff
@@ -93,8 +94,7 @@ export class ArgoCDApiClient implements ArgoCDApi {
     const result = Object.keys(params)
       .filter(key => params[key] !== undefined)
       .filter(
-        key =>
-          this.useNamespacedApps || key === ARGOCD_ANNOTATION_APP_NAMESPACE,
```
alecjacobs5401, line 97: `ARGOCD_ANNOTATION_APP_NAMESPACE` could never be a key here since we aren't passing an annotation, we're passing through the `appNamespace` query param in all cases.

I think the intent is to strip the `appNamespace` query param if the Backstage instance is not opted in, please correct me if that is wrong

alecjacobs5401:
@mwienk do you mind reviewing and validating the original intent as well? Want to make sure I'm not introducing a regression for app namespace support here. Thanks!

mwienk:
The original intent was indeed to remove the query param when not opted in, but only the appNamespace query param should be filtered out, so the change looks good to me.

Xantier (approved):

PR #1215, opened 2024-01-04
Start support for multi-doc transform

Ian-Link-Roadie:
Support for the LoadAll() function when converting YAML to JSON to be transformed by jsonata.  

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.ts
@@ -87,3 +87,73 @@ export function createYamlJSONataTransformAction() {
     },
   });
 }
+
+export function createYamlJSONataTransformAllAction() {
+  return createTemplateAction<{
+    path: string;
+    expression: string;
+    options?: supportedDumpOptions;
+    as?: 'string' | 'object';
+  }>({
+    id: 'roadiehq:utils:jsonata:yaml:transform',
+    description:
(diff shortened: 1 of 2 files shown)


Ian-Link-Roadie:
https://app.shortcut.com/larder/story/18196/support-for-multi-document-yaml-in-yaml-transform-plugin

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.ts
```diff
@@ -87,3 +87,73 @@ export function createYamlJSONataTransformAction() {
...
+      }
+      const sourceFilepath = resolveSafeChildPath(
+        ctx.workspacePath,
+        ctx.input.path,
+      );
+
+      const data = yaml.loadAll(fs.readFileSync(sourceFilepath).toString());
+      const expression = jsonata(ctx.input.expression);
+      data.forEach( (i) => {
+        expression.evaluate(i);
```
Xantier, line 153: This wouldn't work as expected since the returned value from `expression.evaluate` isn't used. 
You could possibly refactor this block to be something like

```typescript
const allResults = data.map( (i) => {
  return expression.evaluate(i);
});
tx.output('result', resultHandler(allResults));
```
(untested, written in browser). 

I'm not certain how the Scaffolder would handle multiple return items as such.

As @punkle mentioned, it might be better to modify the existing one to handle multiyamls via a config option instead of creating a new action altogether.

punkle:
I believe this has been done in another pull request. Closing this for the moment.

PR #1216, opened 2024-01-08
Error when importing detect-indent with scaffolder-backend-module-utils

dlaird-ovo:
This is a patch change to my previous PR - which got missed in our testing.
When running locally all works fine - but when running in a container I see the following error:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /app/node_modules/detect-indent/index.js from /app/node_modules/@roadiehq/scaffolder-backend-module-utils/dist/index.cjs.js not supported.
ovo-backstage-backstage-dev-1  | Instead change the require of index.js in /app/node_modules/@roadiehq/scaffolder-backend-module-utils/dist/index.cjs.js to a dynamic import() which is available in all CommonJS modules.
ovo-backstage-backstage-dev-1  |     at Object.<anonymous> (/app/node_modules/@roadiehq/scaffolder-backend-module-utils/dist/index.cjs.js:14:20)
ovo-backstage-backstage-dev-1  |     at Object.<anonymous> (/app/packages/backend/dist/index.cjs.js:42:36) {
ovo-backstage-backstage-dev-1  |   code: 'ERR_REQUIRE_ESM'
```

This PR changes from `require` to a dynamic import as suggested.  

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-utils/package.json
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-utils/package.json
@@ -38,7 +38,7 @@
     "@backstage/plugin-scaffolder-node": "^0.2.8",
     "adm-zip": "^0.5.9",
     "cross-fetch": "^3.1.4",
-    "detect-indent": "^7.0.1",
+    "detect-indent": "^6.1.0",
     "fs-extra": "^10.0.0",
     "js-yaml": "^4.0.0",
     "jsonata": "^1.8.6",
--- a/.changeset/strange-rice-fry.md
+++ b/.changeset/strange-rice-fry.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
+---
+
+Resolve CommonJS vs ESM Module import use dynamic import)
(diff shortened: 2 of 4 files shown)


dlaird-ovo:
Sadly - I cant get this change to work - runs fine with yarn dev but inside a container I still get the same error - which is disappointing.
I thought dynamic imports were meant to resolve this - but does not appear so.
Investigating.....

Xantier:
These are a bit awkward with Backstage build tools. It may be worth downgrading the library to a version before the change to es-modules only approach

plugins/scaffolder-actions/scaffolder-backend-module-utils/package.json
```diff
@@ -38,7 +38,7 @@
     "@backstage/plugin-scaffolder-node": "^0.2.8",
     "adm-zip": "^0.5.9",
     "cross-fetch": "^3.1.4",
-    "detect-indent": "^7.0.1",
+    "detect-indent": "^6.1.0",
```
Xantier, line 41: oh, sorry. We probably need the modifications to the yarn.lock file into this patch set also

Xantier (approved):

PR #1233, opened 2024-01-25
Support IsDryRun within http-request backend action

electrofelix:
Allow for http methods which should not modify resources to be executed
under dry-run mode. This makes testing templates via dry-run that use
data from http request action output to have their behaviour more
reliably validated.

### DIFF
--- a/.changeset/neat-beers-call.md
+++ b/.changeset/neat-beers-call.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': patch
+---
+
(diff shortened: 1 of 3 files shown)


punkle (approved):

.changeset/neat-beers-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': patch
```
punkle, line 2: Its probably a minor, but not a big deal.

PR #1245, opened 2024-02-09
Update AWS providers 

Xantier:
* Modify Lambda and DDB providers to provide Resource entities
* Add EC2 entity provider to provide EC2 instances.


plugins/backend/catalog-backend-module-aws/src/providers/AWSDynamoDbTableProvider.ts
```diff
@@ -62,7 +62,7 @@ export class AWSDynamoDbTableProvider extends AWSEntityProvider {
 
     const tablePages = paginateListTables(paginatorConfig, {});
 
-    const ddbComponents: ComponentEntity[] = [];
+    let ddbComponents: ResourceEntity[] = [];
```
punkle, line 65: much better

plugins/backend/catalog-backend-module-aws/src/providers/AWSDynamoDbTableProvider.ts
```diff
@@ -89,7 +89,6 @@ export class AWSDynamoDbTableProvider extends AWSEntityProvider {
                 spec: {
                   owner: this.accountId,
                   type: 'dynamo-db-table',
```
punkle, line 91: Do we have a provider to get the rows and map them to a resource with a custom type?

punkle (dismissed):

plugins/backend/catalog-backend-module-aws/src/providers/AWSDynamoDbTableProvider.ts
Xantier: The DDB data provider allows to map rows into entities based on configuration values the user can define

PR #1255, opened 2024-02-14
Align cross-fetch dependency for scaffolder action

ivangonzalezacuna:
This PR is bumping the version of `cross-fetch` to the latest major. This version has been in use for a couple versions of backstage now: https://github.com/backstage/backstage/pull/21229

For some reason, it was still working back then, but since we released the version 1.22 or Backstage, the scaffolder templates using this action would start to fail with a proper reason. Basically, after some investigation, the `fetch` inside the `http` helper method would just get stuck and the `Abort` signal would be triggered after 1 minute.

We were thinking that a routing issue was happening to us, but the same requests with curl and similar tools worked fine. So we tried to use a local version of this plugin but bumping the `cross-fetch` version to align to the other backstage dependencies. Suddenly this got fixed. So I would expect that this could be a potential issue for other people as well. I think it might also make sense to bump all the other dependencies as well, just to make sure this is not happening in other places.

### DIFF
--- a/.changeset/light-guests-begin.md
+++ b/.changeset/light-guests-begin.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': patch
+---
+
+Bump `cross-fetch` to version "^4.0.0"
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/package.json
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/package.json
@@ -35,7 +35,7 @@
     "@backstage/core-app-api": "^1.11.1",
     "@backstage/core-plugin-api": "^1.8.0",
     "@backstage/plugin-scaffolder-backend": "^1.19.1",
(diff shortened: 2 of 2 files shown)


Xantier (commented):
Thanks for the investigation. Were you able to identify the root cause what the version bump fixes?

I think you need to add the modified `yarn.lock` into this PR also before we can ship this

ivangonzalezacuna:
> Thanks for the investigation. Were you able to identify the root cause what the version bump fixes?

Couldn't really find a reason for it. We were debugging and trying different things. We were seeing that many of our testing logs were shown in the UI, but anything written right after the `fetch()` would never be displayed, and after a minute the `AbortError` would be triggered.

> I think you need to add the modified `yarn.lock` into this PR also before we can ship this

I tried that, but a `yarn install` is not modifying the file. I did a `yarn why cross-fetch` and it looks like the cross-fetch version 4.0.0 is already used in many places, probably that's why it didn't change

Xantier (approved):

PR #1261, opened 2024-02-21
implemeting utils fs write action for json prettfied

venkatakurathi:
When using the utility action roadiehq:utils:fs:write, it's noted that any JSON object data is written as a block of text. This observation was made while working on a JSON file to submit a PR, where JSONata operations were performed in my scenario. 

Essentially, this change will ensure that if the input data is in JSON format, its formatting will be preserved when written to a file.
 
Scenario To Produce :
Perfrom a jsonata operation on a json file.
write the output to a new json file.
use the new Json file to create a PR.

Expected outcome : It should be in json format since the input is json.
Current outcome : When the file write operation happens , it is not a prettified/formatted json ...instead its a chunk of text.
image
 in the above scenario , when i used helloworld.json and did some jsonata operation , then wrote that output to helloworld1.json.  When the new json is used as a new file in PR operation , the unprettified json making the whole effort obsolete.      

Another screenshot to showcase the basic print from the newly created json.[post to util action fs write]
image

Post to the implementation of JSON stringify in the fs:write util  action.
image
      
        

- :heavy_check_mark:  Added changeset (run `yarn changeset` in the root)
- :heavy_check_mark:   Screenshots of before and after attached (for UI changes)


Xantier:
Thanks @venkatakurathi ! I believe the same outcome can be achieved by using the serialize action defined to achieve this. So the steps in your template would go something like:
1. Modify the JSON
2. Serialize
3. Write to file.

Could you see if that is something that would alleviate this issue?

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/fs/writeFile.ts
```diff
@@ -56,7 +56,15 @@ export function createWriteFileAction() {
         ctx.input.path,
       );
 
-      fs.outputFileSync(destFilepath, ctx.input.content);
+      let formattedContent = ctx.input.content;
+      try {
```
sblausten, line 60: You could pass a flag to allow users to specify this behaviour to make it non-breaking.

sblausten (commented):
Thanks for this, its definitely an improvement as it reduces the need for extra steps. Just one request in comment

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/fs/writeFile.ts
venkatakurathi: @sblausten i made the processing as optional. With this modification, users can now include a preserveFormatting boolean flag in the input schema. If set to true, the content will be formatted as JSON before writing it to the file, preserving its structure. If set to false or omitted, the content will be written as is without any formatting changes.

venkatakurathi:
> Thanks @venkatakurathi ! I believe the same outcome can be achieved by using the serialize action defined to achieve this. So the steps in your template would go something like:
> 
> 1. Modify the JSON
> 2. Serialize
> 3. Write to file.
> 
> Could you see if that is something that would alleviate this issue?

Thanks @Xantier .

I tried this Approach.
  action: roadiehq:utils:jsonata:json:transform - used the jsonata expression to modify the json object
  action: roadiehq:utils:serialize:json - Passed the ${{ steps.jsonatajson.output.result }} as input data to this action.
ended up with this :
`Invalid input passed to action roadiehq:utils:serialize:json, instance.data is not of a type(s) object
`

I have tried another approach aswell.
action: roadiehq:utils:jsonata:json:transform - used the jsonata expression to modify the json object
action: roadiehq:utils:fs:write : used the above outcome to write it to a new file
action: roadiehq:utils:fs:parse : used json parser on the newly created file

path 1 : create PR post to this . As expected nothing happens to the json. its same as chunk of text.

Path2 :
action: roadiehq:utils:fs:write : used the above outcome to write it to a another json file
encountered the below - `Invalid input passed to action roadiehq:utils:fs:write, instance.content is not of a type(s) string
`
While write file action expects the input to be string and any json parsed outcome when we try to input to the write action it is throwing the below error.

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/fs/writeFile.ts
```diff
@@ -55,8 +66,19 @@ export function createWriteFileAction() {
...
+          formattedContent = JSON.stringify(parsedContent, null, 2);
+        } catch (error) {
+          // Content is not JSON, no need to format
+        }
+      }
+
+      fs.outputFileSync(destFilepath, formattedContent);
 
-      fs.outputFileSync(destFilepath, ctx.input.content);
+      // fs.outputFileSync(destFilepath, ctx.input.content);
```
Xantier, line 81: could you remove this comment?

plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/fs/writeFile.ts
sblausten: That's great thank!

sblausten (approved):

sblausten:
Thanks for seeing this through @venkatakurathi and sorry for slow responses on our side.

PR #1280, opened 2024-03-04
Allow configuration to allow some crude throttling

punkle:
Allow configuration to allow some crude throttling of Okta group membership requests.

### DIFF
--- a/.changeset/poor-jokes-hear.md
+++ b/.changeset/poor-jokes-hear.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-okta': patch
+---
+
+Allow configuration to allow some crude throttling of Okta group membership requests.
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
@@ -37,6 +37,7 @@ import { getOktaGroups } from './getOktaGroups';
 import { getParentGroup } from './getParentGroup';
 import { GroupTree } from './GroupTree';
 import { OktaGroupEntityTransformer, OktaUserEntityTransformer } from './types';
+import chunk from 'lodash/chunk';
 
 /**
  * Provides entities from Okta Org service.
@@ -51,6 +52,7 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
     | { parentKey: string; key?: string }
     | undefined;
   private readonly customAttributesToAnnotationAllowlist: string[];
+  private readonly chunkSize: number;
 
   static fromConfig(
     config: Config,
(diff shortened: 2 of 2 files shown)


.changeset/poor-jokes-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-okta': patch
+---
+
+Allow configuration to allow some crude throttling of Okta group membership requests.
```
Xantier, line 5: Looks to be just chunking here instead of actually throttling the endpoint calls. Is that expected and should we change to changeset to match?

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -168,59 +173,75 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
...
-            await group.listUsers().each(user => {
-              try {
-                const userName = this.userNamingStrategy(user);
-                if (userResources[userName]) {
-                  members.push(userName);
+        for (const chunkOfGroups of chunk(
+          Object.values(oktaGroups),
+          this.chunkSize,
+        )) {
+          const promises = await Promise.allSettled(
```
Xantier, line 180: These are awaited already so actual promise results, maybe a better naming for the array?

sblausten (approved):
Nice

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
```diff
@@ -115,6 +119,7 @@ export class OktaOrgEntityProvider extends OktaEntityProvider {
     this.hierarchyConfig = options.hierarchyConfig;
     this.customAttributesToAnnotationAllowlist =
       options.customAttributesToAnnotationAllowlist || [];
+    this.chunkSize = options.chunkSize || 500;
```
Xantier, line 122: the default seems very high here. How many companies will have over 500 different groups?

.changeset/poor-jokes-hear.md
punkle: I think I have used chunking to effectively limit the number of api calls that are sent to okta, which in my mind is throttling. Am I getting something wrong? Ill add some more detail here.

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
punkle: Ive changed it to 250.

plugins/backend/catalog-backend-module-okta/src/providers/OktaOrgEntityProvider.ts
punkle: Done. thanks.

.changeset/poor-jokes-hear.md
punkle: I have changed the message here. Let me know if it looks better?

Xantier (approved):

PR #1296, opened 2024-03-20
Add AWS org accounts provider

Xantier:


iain-b (approved):
Cool!

The only observation I have is that this appears to run every 5 minutes? no big deal right now but I guess at some point we should consider a) running this on different schedules b) surfacing provider / task schedules to the user and maybe the ability to explicitly run as we could probably poll way less frequently if the user could nudge these as needed (but this is massive tangent obvs).

roadgork (approved):

Xantier:
> The only observation I have is that this appears to run every 5 minutes? no big deal right now but I guess at some point we should consider a) running this on different schedules b) surfacing provider / task schedules to the user and maybe the ability to explicitly run as we could probably poll way less frequently if the user could nudge these as needed (but this is massive tangent obvs).

Ah, yeah. I just added this as is to the example app. The schedule is configured externally from the provider so users can modify this to their liking.

Issue #1300, opened 2024-03-23
[Rag-AI] Error 500 when creating embeddings with AuthMiddleWare enabled


fjudith:
When [calling the endpoint](./plugins/backend/rag-ai-backend#calling-the-endpoint) for embeddings creation, the API returns an 500 HTTP error when the [`authMiddleware` is configured](https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/authenticate-api-requests.md) on the `/catalog` backend router to protect anonymous queries.

## Expected Behavior

Calling the endpoint should work the same way as cases where the `authMiddleware` is not implemented 

## Current Behavior

Below the error message with api request authentication enabled.

```bash
[1] 2024-03-22T22:11:41.763Z backstage info 127.0.0.1 - - [22/Mar/2024:22:11:41 +0000] "GET /api/auth/.well-known/jwks.json HTTP/1.1" 200 754 "-" "-" type=incomingRequest
[1] 2024-03-22T22:11:41.769Z ai info Creating embeddings for source catalog type=plugin
[1] 2024-03-22T22:11:41.771Z backstage info 127.0.0.1 - - [22/Mar/2024:22:11:41 +0000] "GET /api/catalog/entities?filter=kind%3DComponent HTTP/1.1" 401 12 "-" "node-fetch/1.0 (+https://github.com/bitinn/node-fetch)" type=incomingRequest
[1] 2024-03-22T22:11:41.773Z backstage error Request failed with status 500 Request failed with 401 Error type=errorHandler response=[object Response] body=[object Object] cause=Error: Request failed with status 401 Unauthorized, Unauthorized name=ResponseError stack=ResponseError: Request failed with 401 Error
[1]     at Function.fromResponse (/home/fjudith/git/alithya/cloud/toolbox/backstage-app/node_modules/@backstage/errors/src/errors/ResponseError.ts:73:12)
[1]     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
[1]     at CatalogClient.requestRequired (/home/fjudith/git/alithya/cloud/toolbox/backstage-app/node_modules/@backstage/catalog-client/src/CatalogClient.ts:438:13)
[1]     at CatalogClient.getEntities (/home/fjudith/git/alithya/cloud/toolbox/backstage-app/node_modules/@backstage/catalog-client/src/CatalogClient.ts:130:22)
[...]

PR #1307, opened 2024-03-27
Added support for API request authentication.

fjudith:
Relates to issue: https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1300

- Fixes compatiblity with service-to-service auth (ref. https://backstage.io/docs/auth/service-to-service-auth)
- Fixes compatiblity with Pre-1.24..0 API requests authentication method (ref. https://github.com/backstage/backstage/blob/master/contrib/docs/tutorials/authenticate-api-requests.md)


plugins/backend/rag-ai-backend-embeddings-aws/package.json
```diff
@@ -1,10 +1,11 @@
 {
   "name": "@roadiehq/rag-ai-backend-embeddings-aws",
   "description": "The AWS (Bedrock) backend module for the @roadiehq/rag-ai plugin.",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
Xantier, line 4: No need to bump these manually, the changeset and release based on that will do it.

plugins/backend/rag-ai-backend-embeddings-aws/package.json
```diff
@@ -1,10 +1,11 @@
...
   "main": "src/index.ts",
   "types": "src/index.ts",
   "license": "Apache-2.0",
+  "private": true,
```
Xantier, line 8: This shouldn't be here

plugins/backend/rag-ai-backend/README.md
```diff
@@ -302,6 +307,8 @@ export default async function createPlugin({
 
 The `@roadiehq/rag-ai-backend` exposes a router function that can be registered directly into a standard Backstage application.
 
+````typescript
```
Xantier, line 310: this looks funky somehow

plugins/backend/rag-ai-backend-embeddings-aws/package.json
fjudith: Should i revert all versions ?

Xantier: yup, better leave all versions as is. If the package is mentioned in the changeset, it will be bumped automatically

fjudith:
@Xantier i'm not sure why the test is failing i don't see `@backstage/backend-app-api` anywhere in the code.

Should i add the dependency manually ?

Xantier:
> Should i add the dependency manually ?

Yeah, that's probably what is needed here. It is maybe a transitive dep of some of the test packages, so possibly as a dev dep?

fjudith:
> > Should i add the dependency manually ?
> 
> Yeah, that's probably what is needed here. It is maybe a transitive dep of some of the test packages, so possibly as a dev dep?

The package is effectively registered in the dev dependencies already.
Tried to udpated it, but that did'nt fixed the error.

plugins/backend/rag-ai-backend/src/service/RagAiController.ts
```diff
@@ -64,6 +71,9 @@ export class RagAiController {
   }
 
   createEmbeddings = async (req: Request, res: Response) => {
+    const { token } = await this.tokenManager.getToken();
```
Xantier, line 74: These requests are never sent anywhere further so I don't think it's strictly necessary to add the additional headers in this case. Same for the endpoints below

Xantier:
Took a closer look at this. There is a dependency `"@backstage/backend-common": "^0.21.5",` within the `package.json` of `rag-ai` frontend plugin. We need to remove that which should allow the tests to run properly then

fjudith:
> Took a closer look at this. There is a dependency `"@backstage/backend-common": "^0.21.5",` within the `package.json` of `rag-ai` frontend plugin. We need to remove that which should allow the tests to run properly then

Thanks a lot for having taken time to investigate. That makes sense to not have backend plugins as dependencies in frontend plugins.

`yarn test is passing now`

plugins/backend/rag-ai-backend/README.md
```diff
@@ -194,31 +193,38 @@ export default async function createPlugin({
...
-      region: 'eu-central-1',
     },
   });
 
   const model = new Bedrock({
     maxTokens: 4096,
-    model: 'anthropic.claude-instant-v1', // 'amazon.titan-text-express-v1', 'anthropic.claude-v2', 'mistral-xx'
-    region: 'eu-central-1',
+    // model: 'anthropic.claude-instant-v1', // 'amazon.titan-text-express-v1', 'anthropic.claude-v2', 'mistral-xx'*
+    model: 'amazon.titan-text-express-v1', // 'amazon.titan-text-express-v1', 'anthropic.claude-v2', 'mistral-xx'
```
Xantier, line 220:
```suggestion
    model: 'amazon.titan-text-express-v1',
```

These are listed above already

plugins/backend/rag-ai-backend/src/service/RagAiController.ts
```diff
@@ -85,6 +92,9 @@ export class RagAiController {
         message: 'No retrieval pipeline configured for this AI backend. ',
       });
     }
+    const { token } = await this.tokenManager.getToken();
```
Xantier, line 95: Few more of these left here and below. If you remove these, this should be good to go.

Actually the controller doesn't in that case need to token manager to be injected in at all so you could remove that also if you want.

fjudith: Ok it remember that i got errors on the deleteEmbeddings if no token submitted.
Giving a try...

fjudith: I have removed the Token Manager from the controller and the router as well.
Local testing is ok on my end.

Thank you for all theses advices @Xantier

Xantier (approved):
Thanks @fjudith !

Shipping this.

PR #1332, opened 2024-04-11
feat!(plugins/backend/backstage-plugin-argo-cd-backend): terminate argo app operation

vanessap-aa:
Added the ability to terminate an argo application operation both as an its own endpoint, and through the delete argo app and project endpoint.

These changes also created breaking changes as we cleaned up the response to the delete argo app and project route, and added the new functionality of allowing the user to configure how much time to wait between the application delete and project delete. 

Please see changeset for more information.

More details in the 


plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -705,105 +710,151 @@ export class ArgoService implements ArgoServiceApi {
...
+        terminateOperationDetails = {
+          status: 'failed',
+          argoResponse: terminateOperationResp,
+          message: 'application not found',
+        };
+      } else if (terminateOperationResp.statusCode === 200) {
+        terminateOperationDetails = {
+          status: 'success',
+          argoResponse: terminateOperationResp,
+          message: 'application is current operation terminated',
```
Xantier, line 741: This wording could use few tweaks. Maybe `application ${argoAppName} terminated` or something similar?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -951,24 +1002,38 @@ export class ArgoService implements ArgoServiceApi {
...
+  // @see https://cd.apps.argoproj.io/swagger-ui#operation/ApplicationService_List
   async getArgoApplicationInfo({
     argoApplicationName,
     argoInstanceName,
+    baseUrl,
+    argoToken,
   }: {
     argoApplicationName: string;
-    argoInstanceName: string;
+    argoInstanceName?: string;
```
Xantier, line 1013: Any reason to make these optional if they are actually required?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -978,16 +1043,90 @@ export class ArgoService implements ArgoServiceApi {
...
     let statusText: string = '';
+    try {
+      const response: GetArgoApplicationFetchResponse = (await fetch(
+        `${url}/api/v1/applications/${argoApplicationName}`,
+        options,
+      )) as GetArgoApplicationFetchResponse;
+
+      statusText = response.statusText;
+
+      if (response.status === 200) {
```
Xantier, line 1054: This doesn't look to do anything since the return outside of this if clause returns the same response

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -978,16 +1043,90 @@ export class ArgoService implements ArgoServiceApi {
...
+
+  // @see https://cd.apps.argoproj.io/swagger-ui#operation/ApplicationService_TerminateOperation
+  async terminateArgoAppOperation({
+    argoAppName,
+    argoInstanceName,
+    baseUrl,
+    argoToken,
+  }: {
+    argoAppName: string;
+    argoInstanceName?: string;
```
Xantier, line 1078: These again can be non-optional if they are actually required

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
```diff
@@ -368,11 +368,14 @@ export function createRouter({
     async (request, response) => {
```
Xantier, line 368: This looks like it is still leaving the old endpoint in place and adding an additional query param to introduce new functionality. Because of that this looks like that this shouldn't be a major bump.

Am I looking at the wrong thing here? 
Are we able to make these changes without introducing breakages?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
vanessap-aa: I will fix the type to be accurate.

vanessap-aa: great catch!

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
vanessap-aa: I will fix the type to be accurate.  Thank you!

plugins/backend/backstage-plugin-argo-cd-backend/src/service/router.ts
vanessap-aa: correct, the query parameter alone does not introduce a breaking change, however, we did add breaking changes to this PR when changing the response object. The change is explained in the changeset file.  

It would be more accurate to say that this pr introduces a new feature of terminating operations, AND cleans up the delete argo application and project process to more accurately explain the results of the delete calls we are making. Before this, there was poor error handling.

Xantier: Cool. Is there a corresponding modification to the frontend portion of this plugin that is needed for it to work with this newer response shape?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
vanessap-aa: Hello, I have been meaning to respond. I know it looks silly, but without this, typescript does not behave correctly. I have attached some screen shots of when I do not separate the response. 

The desired behavior of the type should show that when the api returns 200 we should get the metadata field, and all other response codes should contain a different data structure.

This photo below shows the type behaving as expected, while separating the returns like we are in the code currently.
Screenshot 2024-04-22 at 12 41 28 PM

This photo below shows the type NOT behaving as expected when we remove the separation. 
Screenshot 2024-04-22 at 12 43 49 PM


Please let me know If you understand that keeping the returns like this helps with the types behaving as the developer would expect.

Kindly.

Xantier (approved):

PR #1341, opened 2024-04-15
Migrate catalog-backend-module-okta to new backend

clementblaise:
- Add scheduler
- Add connect method
- (Breaking) Update config from array to object
- Rename provider-name
- Add alpha module

### DIFF
--- a/packages/backend/src/plugins/catalog.ts
+++ b/packages/backend/src/plugins/catalog.ts
@@ -50,7 +50,7 @@ export default async function createPlugin(
     ...env,
     groupNamingStrategy: 'kebab-case-name',
     userNamingStrategy: 'strip-domain-email',
-  });
+  })[0];
   builder.addEntityProvider(orgProvider);
   providers.push(orgProvider);
 
--- a/plugins/backend/catalog-backend-module-okta/README.md
+++ b/plugins/backend/catalog-backend-module-okta/README.md
@@ -15,7 +15,23 @@ You will need to configure your okta credentials in the `app-config.yaml`.
 catalog:
   providers:
     okta:
-      - orgUrl: 'https://tenant.okta.com'
+      orgUrl: 'https://tenant.okta.com'
+      token: ${OKTA_TOKEN}
+```
+
+# Multiple Configs
+
+You can define multiple config
+
+```yaml
+catalog:
+  providers:
+    okta:
+      org-one:
+        orgUrl: 'https://tenant-one.okta.com'
+        token: ${OKTA_TOKEN}
+      org-two:
+        orgUrl: 'https://tenant-two.okta.com'
         token: ${OKTA_TOKEN}
 ```
 
@@ -27,11 +43,11 @@ catalog:
 catalog:
   providers:
     okta:
-      - orgUrl: 'https://tenant.okta.com'
-        oauth:
-          clientId: ${OKTA_OAUTH_CLIENT_ID},
-          keyId: ${OKTA_OAUTH_KEY_ID},
-          privateKey: ${OKTA_OAUTH_PRIVATE_KEY},
+      orgUrl: 'https://tenant.okta.com'
+      oauth:
+        clientId: ${OKTA_OAUTH_CLIENT_ID},
+        keyId: ${OKTA_OAUTH_KEY_ID},
+        privateKey: ${OKTA_OAUTH_PRIVATE_KEY},
 ```
 
 Note: `keyId` is optional but _must_ be passed wen using a PEM as the `privateKey`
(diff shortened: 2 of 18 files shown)


packages/backend/src/plugins/catalog.ts
```diff
@@ -50,7 +50,7 @@ export default async function createPlugin(
     ...env,
     groupNamingStrategy: 'kebab-case-name',
     userNamingStrategy: 'strip-domain-email',
-  });
+  })[0];
```
punkle, line 53: why only add the first one?

plugins/backend/catalog-backend-module-okta/README.md
```diff
@@ -44,10 +60,10 @@ The provider allows configuring Okta search filtering for users and groups. See
 catalog:
   providers:
     okta:
-      - orgUrl: 'https://tenant.okta.com'
-        token: ${OKTA_TOKEN}
-        userFilter: profile.department eq "engineering"
-        groupFilter: profile.name eq "Everyone"
+      orgUrl: 'https://tenant.okta.com'
```
punkle, line 63: Does this mean we can only configure one okta provider? Surely the array was better?

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -64,17 +68,30 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
...
-    const accountConfig = getAccountConfig(config);
+    if (!options.schedule && !options.scheduler) {
+      throw new Error('Either schedule or scheduler must be provided.');
+    }
+
+    const accountConfig = getAccountConfig(config, DEFAULT_PROVIDER_ID);
+
+    const taskRunner =
+      options.schedule ??
+      options.scheduler!.createScheduledTaskRunner(accountConfig.schedule!);
```
punkle, line 83: Can we schedule when the startup hook has completed? Is that possible with the new backend system?

packages/backend/src/plugins/catalog.ts
clementblaise: I am not aware of the account being configured, the [app-config](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/app-config.yaml) does not have any value for `catalog.providers.okta`

plugins/backend/catalog-backend-module-okta/README.md
clementblaise: No, I updated the [documentation for having multiple account configs](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/5957568ae41ae665d28a0e96089d4c7be61ac80b/plugins/backend/catalog-backend-module-okta/README.md?plain=1#L22-L35). Having an object allows to have proper naming of the catalog location, which is needed to have a scheduler per config.
If it's not under a provider id will use the `default`. It is being done similarly in other catalog providers:
- [catalog-backend-module-msgraph](https://github.com/backstage/backstage/blob/4913addc6531826eee1a6756cef83bb63d3d48b6/plugins/catalog-backend-module-msgraph/src/microsoftGraph/config.ts)
- [catalog-backend-module-github](https://github.com/backstage/backstage/blob/8960902a4ea002d0a44c376f071048b0ef90053f/plugins/catalog-backend-module-github/src/providers/GithubEntityProviderConfig.ts)

plugins/backend/catalog-backend-module-okta/src/module/oktaCatalogModule.ts
```diff
@@ -0,0 +1,105 @@
...
+    env.registerInit({
+      deps: {
+        catalog: catalogProcessingExtensionPoint,
+        config: coreServices.rootConfig,
+        logger: coreServices.logger,
+        scheduler: coreServices.scheduler,
+      },
+      async init({ catalog, config, logger, scheduler }) {
+        catalog.addEntityProvider(
+          OktaOrgEntityProvider.fromConfig(config, {
```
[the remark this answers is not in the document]
clementblaise: I don't think it should be configured as a default; otherwise, they would all run in parallel. If a user wants to use either OktaGroup or OktaUser they can still configure the extension point as [documented](https://backstage.io/docs/backend-system/building-backends/migrating/#other-catalog-extensions)

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
clementblaise: Not a hook but it's configurable: https://github.com/backstage/backstage/blob/ddd67399de0c3ef2e847af90f8e1636f122d972a/packages/backend-tasks/src/tasks/types.ts#L126

clementblaise:
Could you have another look @punkle ?

Xantier:
supereseded by https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1637

PR #1355, opened 2024-04-23
Fix Jira requests for new authentication system

ivangonzalezacuna:
In the latest releases the authentication system has been updated and improved a lot. However, this is now breaking the Jira plugin. In our instance, we would be able to get the project information, but the request against the "search" would fail with a 403.

To fix it, I've changed the old implementation using `fetch` and `identityApi` to inject the token in the headers with directly the `fetchApi`, which does it automatically for you and is the recommended way upstream. Once this has been done, the error 403 obtained in the "search" request is gone.

Also updated the tests to inject the new fetchApi instead of the identityApi


ivangonzalezacuna:
Not sure what is failing in the tests. The implementation should not change at all. It seems like the mock server is working but no data is returned for some reason

kissmikijr:
Isn't it because of the fetchApi mock in the test? It is just a plain jest.fn(). But the code expect it to have a property called ok. Currently it errors with this: 
```
Cannot read properties of undefined (reading 'ok')
```

ivangonzalezacuna:
> Isn't it because of the fetchApi mock in the test? It is just a plain jest.fn(). But the code expect it to have a property called ok. Currently it errors with this:
> 
> ```
> Cannot read properties of undefined (reading 'ok')
> ```

That might be it. Let me try it out real quick

ivangonzalezacuna:
Should be fixed now. The tests were passing locally for me, so I expect the pipeline to pass now. Using some other internal backstage things now, so the tests are a little cleaner now :)

kissmikijr (approved):

kissmikijr:
Thank you and big kudos to you for bringing the tests up to the current standards!

PR #1381, opened 2024-05-07
AWS Providers - Allow matching for exact owner entity ref generation

sblausten:
Enables the option to do matching for owner entity ref generation on resulting resources, if a catalogApi is passed through. 

If no catalogApi is passed the behaviour will stay the same for backwards compatibility and whatever string found in the owner tag will be decorated to the entity owner field.


plugins/backend/catalog-backend-module-aws/src/utils/tags.test.ts
```diff
@@ -58,68 +59,85 @@ describe('labelsFromTags and ownerFromTags', () => {
...
+      expect(result).toBe('group:test/owner1');
+    });
+
+    it('should not return owner from array of tags if the owner is not found in existing groups', () => {
+      const tags = [
+        { Key: 'owner', Value: 'owner3' },
+        { Key: 'tag:two', Value: 'value2' },
+      ];
+      const result = ownerFromTags(tags, undefined, groups);
+      expect(result).toBe('unknown');
```
Xantier, line 87: would it make more sense  to default this to the plain text value instead of unknown in this case?

Xantier (dismissed):

plugins/backend/catalog-backend-module-aws/src/utils/tags.test.ts
sblausten: Yes could do but the owner won't relate to a catalog entity then. Maybe that's fine

Xantier: yeah, I was thinking that the unknown won't likely match catalog either but at least the attempted value could give some information what was the initial wanted value.

sblausten: Yea true. I changed it to that now

Xantier (approved):

PR #1393, opened 2024-05-18
Add matchRegex to `roadiehq:utils:fs:replace` to allow for more complex replacements

alecsammon:
When using  `roadiehq:utils:fs:replace` then you may wish to replace based on a regular expression.

For example we want to create a template that updates the `go.mod` to the latest go version. As we don't know which version the source is using then without a regular expression match we would have to search for every different option.


punkle (approved):
Looks great. I cant find fault. Thank you. 🙏

PR #1413, opened 2024-06-04
Iframe from templated src

alecsammon:
Allow constructing an iframe with a string with multiple replacements.

Current the iframe plugin allows adding a source with a static string, or a reference to an annotation.

This extends the plugin to allow creating a source from a pattern and multiple annotations.
This is useful if you want to point at another tool and include the entity name as a query parameter.
Or link to documentation depending on the language of the component.

### DIFF
--- a/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
+++ b/plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
@@ -21,6 +21,7 @@ import {
   IFrameComponentContentProps,
   IFrameComponentProps,
   IFrameFromAnnotationProps,
+  IFrameFromAnnotationReplacementsProps,
 } from './types';
 import { Content, ContentHeader } from '@backstage/core-components';
 import { configApiRef, useApi } from '@backstage/core-plugin-api';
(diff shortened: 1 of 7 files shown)


plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
```diff
@@ -109,7 +152,14 @@ const IFrameFromSrc = (props: IFrameProps) => {
  * @public
  */
 export const IFrameCard = (props: IFrameComponentProps) => {
-  const { src, srcFromAnnotation, height, width, title } = props;
+  const {
+    src,
+    srcFromAnnotation,
+    srcWithAnnotationReplacements,
```
Xantier, line 158: Let's call this something like templatedSrc or srcWithTemplate or similar I think

Xantier (commented):
A naming thing and tests look to fail but other than those, this is nearly good to go

plugins/frontend/backstage-plugin-iframe/src/components/IFrameComponent.tsx
alecsammon: Agree! Thanks.

Updated to `templatedSrc`

alecsammon:
Thanks @Xantier  - I've update the naming as suggested, and fixed a problem with the version of `@backstage/catalog-model`. 

Build is now 🟢 

When you have time to review again that would be greatly appreciated.

Xantier (approved):
Nice!

alecsammon:
Thank you ❤️

PR #1451, opened 2024-06-26
sanitize the aws label values better

punkle:
sanitize the aws label values better

### DIFF
--- a/.changeset/tidy-ghosts-care.md
+++ b/.changeset/tidy-ghosts-care.md
@@ -1,5 +1,5 @@
 ---
-'@roadiehq/catalog-backend-module-aws': minor
+'@roadiehq/catalog-backend-module-aws': patch
 ---
 
 Label values are sanitized a bit better.
--- a/.changeset/tidy-ghosts-care.md
+++ b/.changeset/tidy-ghosts-care.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-aws': minor
+---
+
+Label values are sanitized a bit better.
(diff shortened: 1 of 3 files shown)


.changeset/tidy-ghosts-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-aws': minor
```
Xantier, line 2: patch is enough

Xantier (dismissed):

plugins/backend/catalog-backend-module-aws/src/utils/tags.ts
```diff
@@ -46,7 +46,9 @@ export const labelsFromTags = (tags?: Tag[] | Record<string, string>) => {
       .reduce((acc: Record<string, string>, tag) => {
         if (tag.Key && tag.Value) {
           const key = tag.Key.replaceAll(':', '_').replaceAll('/', '-');
-          acc[key] = tag.Value.replaceAll('/', '-').substring(0, 63);
+          acc[key] = tag.Value.replaceAll('/', '-')
```
iain-b, line 49: I really wonder if we should be adding these as labels at all? These changes materially affect the value and they might not be meaningful / usable without them. We could perhaps use tags? e.g. tags: ['keyA=valueA'] ?

punkle: Im not sure tbh.

Xantier (approved):
I'm pro leaving this as labels for now. This will allow more complex entity filtering options if there are need to build such

PR #1474, opened 2024-07-10
Add more sanitization of label values

Xantier:


plugins/backend/catalog-backend-module-aws/src/utils/tags.ts
```diff
@@ -71,8 +71,11 @@ export const labelsFromTags = (tags?: Tag[] | Record<string, string>) => {
     )
     .reduce((acc: Record<string, string>, [key, value]) => {
       if (key && value) {
-        const k = key.replaceAll(':', '_').replaceAll('/', '-');
-        acc[k] = value.replaceAll('/', '-').substring(0, 63);
+        let k = key.replaceAll(':', '_').replaceAll('/', '-');
```
punkle, line 74: I do wonder if [slugify](https://www.npmjs.com/package/slugify) would drop in here.

punkle (approved):

.changeset/tasty-yaks-agree.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/catalog-backend-module-aws': patch
+---
+
+Add more sanitization to AWS tags to labels functionality
```
punkle, line 5: We should add a test for this case.

Issue #1490, opened 2024-07-21
Buildkite plugin: support option for dynamically detecting the default branch and only showing its builds


mdb:
PR #1460 added support to the Buildkite plugin for a `buildkite.com/branch` annotation. Additionally, it would be useful to support the ability to configure the plugin to only display default branch builds, without needing to to hardcode references to the default branch name as an entity YAML annotation. 

## Feature Suggestion

Support for the ability to dynamically determine the default branch name would save users from having to clutter entity YAML with hard-coded, brittle `buildkite.com/branch` annotations.

As a bonus, this would involve adding a `getPipeline` method to the plugin's [BK API client](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/frontend/backstage-plugin-buildkite/src/api/index.ts), which would be useful for users in other contexts. In particular, it would enable the use of the [backstage-plugin-buildkite](https://github.com/RoadieHQ/roadie-backstage-plugins/tree/main/plugins/frontend/backstage-plugin-buildkite)'s API client in developing a [Buildkite CicdStaticsApi implementation](https://github.com/backstage/community-plugins/tree/main/workspaces/cicd-statistics/plugins/cicd-statistics#usage) for use in integrating Buildkite with the Backstage [CI/CD statics plugin](https://github.com/backstage/community-plugins/tree/main/workspaces/cicd-statistics/plugins/cicd-statistics).

## Possible Implementation

Globally, the [BuildkiteBuildsTable](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/frontend/backstage-plugin-buildkite/src/components/BuildKiteBuildsTable/BuildKiteBuildsTable.tsx#L151) could support something like a `onlyDisplayDefaultBranch` property.

Additionally -- or alternatively -- a `buildkite.com/default-branch-only: 'true'` entity YAML annotation could configure the plugin to only display default branch builds on a per-entity basis.

I'm imagining the plugin can continue to support a more specific `buildkite.com/branch` annotation on entity YAML, whose value may take precedence above both `buildkite.com/default-branch-only` as well as a globally-set `onlyDisplayDefaultBranch` property.

PR #1460, opened 2024-06-26
buildkite plugin: optionally filter builds to specific branch

mdb:
This addresses issue #1459 to support the ability to optionally restrict the Buildkite builds table to only display builds of a particular branch. The branch filter is controlled via an entity annotation. For example:

```yaml
metadata:
  annotations:
    'buildkite.com/branch': 'main'
```

#### :heavy_check_mark: Checklist

- [X] Added tests for new functionality and regression tests for bug fixes
- [X] Added changeset (run `yarn changeset` in the root)
- [ ] Screenshots of before and after attached (for UI changes)
- [X] Added or updated documentation (if applicable)

PR #1498, opened 2024-07-25
buildkite plugin: support 'defaultBranchOnly' property

mdb:
This addresses issue #1490 and enables the ability to configure the backstage-plugin-buildkite to only show default branch builds (without needing to hard-code a `buildkite.com/branch` annotation specifying that default branch name).

On a per-entity descriptor basis, the option can be configured via an annotation:

```yaml
metadata:
  annotations:
    'buildkite.com/default-branch-only': 'true'
```

Alternatively, the option can also be configured globally:

```ts
// packages/app/src/components/catalog/EntityPage.tsx export const cicdContent = (
  <EntitySwitch>
    <EntitySwitch.Case if={isBuildkiteAvailable}>
      <EntityBuildkiteContent defaultBranchOnly />
    </EntitySwitch.Case>
    ...
  </EntitySwitch>
);
```

The presence of a `buildkite.com/branch` annotation continues to take
precedence over both a globally-set `defaultBranchOnly` and a
`buildkite.com/default-branch-only` annotation, preserving backwards
compatibility.

Additionally, if the plugin has been configured to only display a
specific branch's builds, the branch name is now rendered in parentheses
beside the pipeline name.

### DIFF
--- a/.changeset/olive-bats-visit.md
+++ b/.changeset/olive-bats-visit.md
@@ -1,0 +1,34 @@
+---
+'@roadiehq/backstage-plugin-buildkite': patch
+---
+
+Address issue #1490 to support the ability to only display default branch builds.
+
+A `defaultBranchOnly` property can be set on the component, configuring the
+component to dynamically detect the entity's default branch and only display
+builds of that branch:
+
+```ts
+// packages/app/src/components/catalog/EntityPage.tsx
+export const cicdContent = (
+  <EntitySwitch>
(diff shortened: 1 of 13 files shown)


mdb:
👋 Hi @kissmikijr - This is a followup to PR #1460 , which you reviewed and merged 2 weeks ago. Would you be willing to offer a review here, too? Thanks!

Xantier (approved):
Nice! Thanks for the addition

Issue #1475, opened 2024-07-10
 Issue with Token Retrieval in http:backstage:request Action - failing with JWTClaimValidationFailed


davormilutinovic:
**Expected Behavior**
The http:backstage:request action should use the initiator's credentials to retrieve the token, ensuring proper authentication when making HTTP requests.

**Current Behavior**
The http:backstage:request action uses old ctx.secrets?.backstageToken or ctx.secrets.backstageToken to retrieve the token, which leads to an authentication error with the message "Invalid plugin token; caused by JWTClaimValidationFailed: unexpected 'aud' claim value".

link to code line :https://github.com/RoadieHQ/roadie-backstage-plugins/blob/c62ec52707b3be39c40d240d299dba00995a3243/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts#L120

**Steps to Reproduce**
Configure a template in Backstage that uses the http:backstage:request action to make an authenticated HTTP request.
Execute the template to trigger the action.
Observe the authentication error in the logs.

**Possible Solution**

Modify the http:backstage:request action to use ctx.getInitiatorCredentials().token instead of ctx.secrets?.backstageToken 

**Quick fix**

```js
// Original line
const token = ctx.secrets?.backstageToken; 

// Updated lines
const credentials = await ctx.getInitiatorCredentials(); 

// @ts-expect-error
const token = credentials.token;
```

**Context**

This issue prevents us from properly authenticating HTTP requests within the Backstage scaffolder, causing our pipeline creation process to fail with authentication errors. We are trying to automate the creation of SonarQube pipelines in Azure, and this bug is a blocker for our workflow.

**Your Environment**
yarn: 1.22.21
cli:  0.26.6 (installed)
backstage:  1.27.6

Dependencies:
 "@roadiehq/scaffolder-backend-module-http-request": "^4.3.2",
  @backstage/app-defaults                                          1.5.5
  @backstage/backend-app-api                                       0.7.5
  @backstage/backend-common                                        0.22.0
  @backstage/backend-defaults                                      0.2.18
  @backstage/backend-dev-utils                                     0.1.4
  @backstage/backend-openapi-utils                                 0.1.11
  @backstage/backend-plugin-api                                    0.6.21
  @backstage/backend-tasks                                         0.5.26
  @backstage/catalog-client                                        1.6.5
  @backstage/catalog-model                                         1.5.0
[...]

PR #1514, opened 2024-07-29
Fix authentication issues with the http-requests against the proxy

ivangonzalezacuna:
The handler was using the old authentication setup instead of the new one using the new auth service. Let's use this instead and always set the token if defined. This is following the same idea as other upstream actions, such as the one to register a new component: https://github.com/backstage/backstage/blob/e6c52e7e7ec21ceb98cfbaad55c84b3e3683649f/plugins/scaffolder-backend/src/scaffolder/actions/builtin/catalog/fetch.ts#L93

In our setup, we needed to set the `credentials: dangerously-allow-unauthenticated` temporarily. I've tested it without them again and the requests weren't returning the 401 any more.

This should be fixing https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1475, and I followed the idea in https://github.com/backstage/backstage/issues/25096#issuecomment-2157710780

### DIFF
--- a/.changeset/smart-peas-mate.md
+++ b/.changeset/smart-peas-mate.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': patch
+---
+
+Fix authentication to the proxy in the scaffolder action
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
@@ -14,19 +14,17 @@
  * limitations under the License.
  */
 
-import { createTemplateAction } from '@backstage/plugin-scaffolder-backend';
-import {
-  generateBackstageUrl,
-  http,
-  getObjFieldCaseInsensitively,
-} from './helpers';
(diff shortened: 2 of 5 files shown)


davormilutinovic:
Hi. For some reason your changes are still not working for me.
I have used your fork and there was an error "...There was an issue with your request. Status code: 401 Response body: {"error":{"name":"AuthenticationError","message":"Invalid plugin token; caused by JWTClaimValidationFailed: unexpected \"aud\" claim value","cause":{"code":"ERR_JWT_CLAIM_VALIDATION_FAILED..."

After I reverted to 
"const credentials = await ctx.getInitiatorCredentials(); 

// @ts-expect-error
const token = credentials.token;"

It has start working again?

Nitinvonage:
I am still getting the same issue with latest version...

```
@roadiehq/scaffolder-backend-module-http-request@5.3.0
backstage version - 1.35.1
error i am getting while making a post call from template is
```

error: There was an issue with your request. Status code: 401 Response body: {"error":**{"name":"AuthenticationError","message":"Failed plugin token verification"}**,"request":{"method":"POST","url":"/service/api/prd?message=test-user"},"response":{"statusCode":401}}52025-02-21T10:33:02.158Z Error: Unable to complete request

PR #1533, opened 2024-08-06
Handle cases where S3 buckets don't have a tag set

Xantier:


plugins/backend/catalog-backend-module-aws/src/providers/AWSS3BucketProvider.ts
```diff
@@ -94,8 +94,15 @@ export class AWSS3BucketProvider extends AWSEntityProvider {
...
-        const tagsResponse = await s3.getBucketTagging({ Bucket: bucket.Name });
-        const tags = tagsResponse?.TagSet ?? [];
+        let tags: Tag[] = [];
+        try {
+          const tagsResponse = await s3.getBucketTagging({
+            Bucket: bucket.Name,
+          });
+          tags = tagsResponse?.TagSet ?? [];
+        } catch (e) {
+          this.logger.debug('No tags found for bucket');
```
punkle, line 104: hmm. I guess you dont know if this is the cause of the exception. I would use: failed to retrieve any tags for the bucket. Also no harm in putting in a few metadata items, like the bucket name maybe.

punkle (dismissed):

plugins/backend/catalog-backend-module-aws/src/providers/AWSS3BucketProvider.ts
Xantier: Thanks, updated

kissmikijr (approved):

Issue #1535, opened 2024-08-06
scaffolder-backend-module-utils is not passing options to npm yaml package correctly


timdoherty:
See [this issue](https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1301) that was closed as stale. The underlying [yaml library was changed](https://github.com/RoadieHQ/roadie-backstage-plugins/commit/3d55eb2fd4131c670405d5370b9bcb405ca3c168) from `js-yaml` to `yaml` but the interfaces are different and options are being passed incorrectly to `yaml` since the `yamlOptionsSchema` is unchanged. 

Specifically, this is breaking `noArrayIndent` since `yaml` expects this to be passed as the option `{ indentSeq: false }`

## Expected Behavior

Should expose `indentSeq` via `yamlOptionsSchema` such that providing `indentSeq: false` results in:

```yaml
my_array:
- item 1
- item 2
- item 3
```

## Current Behavior

`noArrayIndent` is ignored, since it's not a valid option for `yaml`, and the default `indent` value of `2` is used.

```yaml
my_array:
  - item 1
  - item 2
  - item 3
```

## Steps to Reproduce

Use the `roadiehq:utils:jsonata:yaml:transform` action when scaffolding a template to transform some yaml array with the `noArrayIndent` option set to true. It will still indent the entries.

## Possible Solution

- Update [yamlOptionsSchema](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/53ad92fce879517057337ba59bdf240811f6a341/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/types.ts#L21) `yamlOptionsSchema` to use the correct options from [yaml](https://eemeli.org/yaml/#options)
- Update [this assertion](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/53ad92fce879517057337ba59bdf240811f6a341/plugins/scaffolder-actions/scaffolder-backend-module-utils/src/actions/jsonata/yaml.test.ts#L116) or add another test to validate the output vs. just the options passed

## Context

We use `roadiehq:utils:jsonata:yaml:transform` to open PRs against another team's repository, and this bug violates their YAML style guide, causing organizational friction and rework.

## Your Environment

```json
    "@backstage/backend-app-api": "^0.8.0",
    "@backstage/backend-common": "^0.23.3",
    "@backstage/backend-defaults": "^0.4.1",
    "@backstage/backend-plugin-api": "^0.7.0",
[...]

PR #1241, opened 2024-02-06
Swap YAML library in Scaffolder actions to a better one

JoaoMartins51:
https://app.shortcut.com/larder/story/18744/globalization-partners-swap-yaml-library-in-scaffolder-actions-to-a-better-one

#### :heavy_check_mark: Checklist

- [ ] Added tests for new functionality and regression tests for bug fixes
- [ ] Added changeset (run `yarn changeset` in the root)
- [ ] Screenshots of before and after attached (for UI changes)
- [ ] Added or updated documentation (if applicable)

PR #1542, opened 2024-08-09
UIS-1326: Scaffolder backend module utils - patches yamlOptionsSchema to match 'yaml' options

timdoherty:
Patches [1535](https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1535)

[This PR](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1241) swapped out `ys-yaml` for `yaml` and updated the options type, but neglected to update the `yamlOptionsSchema` accordingly. The result is that most of the options exposed by the plugin are ignore by `yaml` and the plugin is not functioning as expected.

This PR updates the `yamlOptionsSchema` to match the options expected by `yaml`, restoring correct functionality.  


plugins/scaffolder-actions/scaffolder-backend-module-utils/src/types.ts
```diff
@@ -23,59 +23,89 @@ export const yamlOptionsSchema = {
   description: 'YAML stringify options',
   type: 'object',
   properties: {
-    indent: {
-      description: '(default: 2) - indentation width to use (in spaces)',
-      type: 'number',
+    blockQuote: {
```
timdoherty, line 26: `commentString` was omitted in the [original PR](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1241/commits/42db6b8f18c5e1b1ba3e685545b63118ed54ef4f#diff-0c7f5cb766fd6b51674c33ddddd04f5d7cc7aa0a906a690c4f9854d34c1eca19R19) - I'm not sure the reasoning but I've respected that decision by omitting it from the `yamlOptionsSchema` and updated test inputs.

.changeset/clean-seals-sip.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/scaffolder-backend-module-utils': patch
```
iain-b, line 2: I believe this is actually a major change even though it's correcting the types? (feel free to disagree) e.g. what currently compiles (but doesn't work as expected) will not now compile? I understand the PR which switched the library was also a patch but that seems to have been a mistake

iain-b (commented):
Thanks so much for the contribution! just one comment around versioning?

.changeset/clean-seals-sip.md
timdoherty: Agreed, I think the original PR should have been a breaking change, I'm happy to update accordingly.

Can I just clarify this statement in the contributing doc:

> Select impact of change that you're introducing, using minor for breaking changes and patch otherwise.

Is this still the case, all breaking changes are *minor* version bumps? Any reason not to use semantic versioning?

iain-b: There's a convention (not sure how widely followed, it was new to me) that pre version 1 (e.g. 0.2.3) that minors can be used for breaking changes as you'd likely want to make a bunch of breaking changes prior to a formal release. 

I just checked Backstage and they've qualified this line:
```
Select impact of the change you're introducing. If the package you are changing is at version 0.x, use minor for breaking changes and patch otherwise. If the package is at 1.0.0 or higher, use major for breaking changes, minor for backwards compatible API changes, and patch otherwise. See the [Semantic Versioning specification](https://semver.org/#semantic-versioning-specification-semver) for more details.
```

Frankly, I'd be inclined to always follow semver as users are likely going to pull in v0 packages anyway. I'll discuss internally we my team and we'll update this doc.

iain-b (approved):

.changeset/clean-seals-sip.md
timdoherty: Done, both `major` changeset and update to CONTRIBUTING.md

iain-b (approved):

PR #1543, opened 2024-08-13
fix(jira): jira plugin no longer makes unnecessary separate api calls to get issue counts

roope-t:
Before this change, Jira tile was making API calls to fetch project information (includes issue types), all issues of the project, and issue counts for each issue type. This caused as many unnecessary API calls as there were issue types in the project (in our case ~90 unnecessary calls).

Because the issues themselves contain information about the issue type, we can just get the counts from all issues and project information (with issue types).

### DIFF
--- a/plugins/frontend/backstage-plugin-jira/src/api/index.ts
+++ b/plugins/frontend/backstage-plugin-jira/src/api/index.ts
@@ -21,8 +21,8 @@ import {
   FetchApi,
 } from '@backstage/core-plugin-api';
 import {
-  IssueCountResult,
-  IssueCountSearchParams,
+  IssuesResult,
+  IssuesResponse,
   IssuesCounter,
   IssueType,
   Project,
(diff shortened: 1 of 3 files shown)


plugins/frontend/backstage-plugin-jira/.changeset/config.json
```diff
@@ -0,0 +1,11 @@
+{
```
Xantier, line 1: You can drop this file. We have a config already present in the repo root.

plugins/frontend/backstage-plugin-jira/src/api/index.ts
```diff
@@ -241,66 +190,39 @@ export class JiraAPI {
...
+
+    const foundIssues = await this.getIssuesPaged({
+      apiUrl,
+      projectKey,
+      component,
+      label,
+      statusesNames,
+    });
+
+    const issuesCounter: IssuesCounter[] = issuesTypes.map(issueType => {
```
Xantier, line 211: You might be better off using `lodash` `groupBy` or reducing it manually in here to avoid having the need to loop through the issues multiple times. I don't know how big the `issueTypes` collection grows, so it could be fine on smaller amounts.

roope-t: Good point, I decided to just "merge" the issueTypes and issueCounter loops, as there was really no reason do have them separate.

Xantier (approved):

PR #1559, opened 2024-08-20
feat(ai): add support for chat completion models

benjidotsh:
This PR adds support for the newer [chat completion models](https://js.langchain.com/v0.2/docs/concepts/#chat-models), in addition to the currently supported [text completion models](https://js.langchain.com/v0.2/docs/concepts/#llms).

This allows you to use Langchain models like [`BedrockChat`](https://js.langchain.com/v0.2/docs/integrations/chat/bedrock), which in turn allow you to use models like Meta's Llama 3 70B:

```typescript
const model = new BedrockChat({
  model: 'meta.llama3-70b-instruct-v1:0',
  region: 'us-east-1',
  credentials: credProvider.sdkCredentialProvider,
});
```

### DIFF
--- a/plugins/backend/rag-ai-backend/src/service/LlmService.ts
+++ b/plugins/backend/rag-ai-backend/src/service/LlmService.ts
@@ -14,13 +14,14 @@
  * limitations under the License.
  */
 import { BaseLLM } from '@langchain/core/language_models/llms';
+import { BaseChatModel } from '@langchain/core/language_models/chat_models';
 import { EmbeddingDoc } from '@roadiehq/rag-ai-node';
 import { Logger } from 'winston';
 import { createPromptTemplates } from './prompts';
 
 export class LlmService {
   private readonly logger: Logger;
-  private readonly model: BaseLLM;
+  private readonly model: BaseLLM | BaseChatModel;
   private readonly prompts: {
     prefixPrompt: (embedding: string) => string;
     suffixPrompt: (input: string) => string;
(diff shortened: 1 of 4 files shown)


plugins/backend/rag-ai-backend/src/service/LlmService.ts
```diff
@@ -51,6 +52,7 @@ export class LlmService {
     const prompt = `Human:\n${this.prompts.prefixPrompt(
       promptEmbeddings,
     )}\n ---\n${this.prompts.suffixPrompt(query)}\nAssistant:`;
-    return await this.model.invoke(prompt);
+    const response = await this.model.invoke(prompt);
+    return typeof response === 'string' ? response : response.content;
```
Xantier, line 56: Are we able to determine the type of response shape more specifically here?

Xantier (commented):
Nice! One question that could maybe be improved .

The next stages I guess for this would be to extend the functionality to be able to embed historicals into the context if a proper back and forth is a wanted way to go about.

plugins/backend/rag-ai-backend/src/service/LlmService.ts
benjidotsh: Yes: `this.model._modelType()` returns `base_chat_model`, but then we lose TypeScript's automatic type inference. That's why I didn't end up using it.

Xantier: Aight, grand. Let's ship

Xantier (approved):

PR #1561, opened 2024-08-21
feat(ai): add support for streaming responses

benjidotsh:
This PR adds support for streaming responses and bumps all `@langchain` packages to the latest version.

[According to the Langchain docs](https://js.langchain.com/v0.2/docs/concepts/#stream), this implementation should still work for models that don't support streaming by streaming a single chunk, but I didn't test this, so some more testing might be useful.

Since the implementation of the query endpoint in both the frontend and backend has drastically changed and makes use of server-sent events, this PR definitely contains a couple of **breaking changes** in the sense that an older version of the frontend won't work with this version of the backend and vice versa. Other than that, the user doesn't have to change anything. If the model supports streaming, the response will be streamed. If not, it should still work as well.

There is 1 TODO in the `RagAiController` because adding the `'Content-Type': 'text/event-stream'` seems to result in buffered responses in the frontend, although that seems to be a recommended header for server-sent events. Not sure how important that is.

https://github.com/user-attachments/assets/ef9ffae7-a85c-4ba2-8b47-b4c4a6d17eda


Xantier:
Thanks @benjidotsh! I will be getting deeper into looking at this later this week (friday most likely). But before that: I believe it is time with this PR to bump the version numbers to 1.0.0. It's best practice to iterate above 1.x so we can manage semver better. Could you change the changeset to accommodate that with a `major` bump.

benjidotsh:
@Xantier the frontend is already `1.0.0`, so only the backend I suppose?

Xantier:
> @Xantier the frontend is already 1.0.0, so only the backend I suppose?

Indeed, yup

fjudith:
@benjidotsh Did you tested this feature using Amazon Bedrock ro OpenAI ?

benjidotsh:
@fjudith no worries at all!

That's really strange... Are you sure you ran `yarn install` after checking out this branch? And can you give me the initialization code of your model? I'll try to reproduce.

fjudith:
> @fjudith no worries at all!
> 
> That's really strange... Are you sure you ran `yarn install` after checking out this branch? And can you give me the initialization code of your model? I'll try to reproduce.

Sure: ! <https://github.com/testruction/roadie-backstage-plugins/blob/ai/streaming/packages/backend/src/plugins/ai.ts>

benjidotsh:
@fjudith ah, I see it! You're using `amazon.titan-embed-text-v2:0`, which is a model for creating embeddings and not for querying. For embeddings, you need to configure the model in `app-config.yaml` like so:
```typescript
ai:
  embeddings:
    bedrock:
      modelName: 'amazon.titan-embed-text-v2:0'
```

fjudith:
> @fjudith ah, I see it! You're using `amazon.titan-embed-text-v2:0`, which is a model for creating embeddings and not for querying. For embeddings, you need to configure the model in `app-config.yaml` like so:
> 
> ```ts
> ai:
>   embeddings:
>     bedrock:
>       modelName: 'amazon.titan-embed-text-v2:0'
> ```
> 
> I was already confused when you mentioned "any model supporting embedding"... 😉

Ohoh! i understand here that the model in the `app-config.yaml` needs to be different from the one in the `ais.ts`.

benjidotsh:
@fjudith indeed! The one in `ai.ts` is the model you use for querying and the one in `app-config.yaml` is the one you use for creating embeddings.

Did this work for you before, on the current release?

fjudith:
> @fjudith indeed! The one in `ai.ts` is the model you use for querying and the one in `app-config.yaml` is the one you use for creating embeddings.
> 
> Did this work for you before, on the current release?

I'm pretty sure that i introduced mysrlf the configuration error when the Bedrock opened in `ca-central-1` en of June, before i was just copy-pasting the documented configuration.

fjudith (approved):

benjidotsh:
@fjudith no worries! Thanks for testing it out! 😉

.changeset/stupid-rings-buy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@roadiehq/rag-ai-backend': major
+'@roadiehq/rag-ai': minor
```
Xantier, line 3: This can be minor since we are starting from 1.0.0 already

Xantier: Ah, no wait. It's actually changing the interface type. Fine to keep as major

Xantier (approved):
Shipping. I will bump the underlying modules to 1.0.0. separately after dependency upgrades etc.

PR #1566, opened 2024-08-23
add default entity filter to datadog plugin (new FE sys only)

alex-mckay:
Follow up from FE system compatibility work, ensure the Datadog plugin, when using the [new FE system](https://backstage.io/docs/frontend-system/), includes a filter to publish the entity content to components & resources only by default. This can be overridden. Also adds docs about how to do that overriding.

Since the Roadie repo upgraded to backstage 1.30, this PR also updates the syntax for the new frontend system within the datadog plugin according to [these changes](https://backstage.io/docs/frontend-system/architecture/migrations#130). Is a functional no-op/only gets rid of deprecated references. 


alex-mckay:
@Xantier would appreciate a review next week! Small change for portal ease of use, and I added some docs for the new FE specifially :)

Xantier (changes requested):
Thanks. This seems like the wrong approach on this frontend plugin. Let's do these filter via props instead.

alex-mckay:
hey @Xantier! For the new frontend system in particular, declaring entity content or cards to export has [this built in filter parameter](https://github.com/backstage/backstage/blob/v1.26.5/plugins/catalog-react/api-report-alpha.md?plain=1#L43-L45) to control which entity kinds the content/cards will get mounted onto. Trying to pass the control mechanism via props or inputs into this entity card/component instead wouldn't work the components 'above' it in the hierarchy wouldn't know to utilize that input

alex-mckay:
Hey @Xantier - let me know if you have questions about the comment above!

Xantier:
Thanks @alex-mckay, 
I just got back from vacation. This looks ok, though the PR seems to be out of date at the moment

nuritizra:
@karlhaworth :pray:

PR #1616, opened 2024-09-12
Some fixes on migrating the Argocd FE plugin

nuritizra:
Added some fixes while the [docs](https://backstage.io/docs/frontend-system/building-plugins/migrating) were updated to use blueprints.
Also, we don't need the page extension as we only use cards.
In addition, I updated the README file.


Xantier:
Could we at the same time modify the naming convention in exporting the plugin. It looks very unprofessional to be exporting `alpha` functionality and I'm not sure if that's worth doing. Is the frontend system finalized at this stage?

Rugvip:
@Xantier the frontend system is not stable yet and there will be more breaking changes. Whether you want to have it exported via `/alpha` or not is up to you tbh, but you'd ofc need to make sure you forward any breaking changes from the core APIs that affect the external API of your package.

Xantier:
Hmm, thanks @Rugvip. @nuritizra would it make sense to hold off on this until the frontend system gets into a stabler state. Currently this addition looks to be just introducing a maintenance burden.

Rugvip:
@Xantier these changes are just helping out with the migrations from https://backstage.io/docs/frontend-system/architecture/migrations#130. We want to push for adoption of the new frontend system by plugins already now, because it's hard to move to and encourage adoption in apps until we have decent support from plugins.

PR #1628, opened 2024-09-19
refactor(rag-pgvector): use LoggerService instead of winston

kuangp:
Switch to `LoggerService` interface to allow new backend system apps to remove usages of the deprecated `loggerToWinstonLogger` utility while still being backwards compatible with winston 

### DIFF
--- a/.changeset/hip-lamps-cheat.md
+++ b/.changeset/hip-lamps-cheat.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/rag-ai-storage-pgvector': major
+---
+
+Switch from winston to `LoggerService` interface to support new backend system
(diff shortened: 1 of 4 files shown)


Xantier (approved):

.changeset/hip-lamps-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/rag-ai-storage-pgvector': patch
```
Xantier, line 2: I believe this is a breaking change due to interface changes. Let's make it a `major` so we can start following proper semver

kuangp: updated!

PR #1629, opened 2024-09-19
feat(rag-retrieval): support new auth system

kuangp:
Provide backwards compatible support for new auth system and replace winston with `LoggerService` interface

### DIFF
--- a/.changeset/lucky-maps-confess.md
+++ b/.changeset/lucky-maps-confess.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/rag-ai-backend-retrieval-augmenter': major
+---
+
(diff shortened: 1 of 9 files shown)


.changeset/lucky-maps-confess.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/rag-ai-backend-retrieval-augmenter': patch
```
Xantier, line 2: Let's make this major since we are changing interfaces. That way we can get to the mainline versioning strategy at the same time.

kuangp: updated!

Xantier (approved):
Thanks!

PR #1703, opened 2024-10-28
Fix README.md examples for catalog-backend-module-okta

jdnewman85:
Documentation fixes only.
Commas break the auth fields by getting appended to the interpolated values.
Function signature for the group transformer was incorrect with parent and options swapped. See definition: https://github.com/RoadieHQ/roadie-backstage-plugins/blob/0d983a2f5a728122e66b966529fde7860b6377f8/plugins/backend/catalog-backend-module-okta/src/providers/types.ts#L25

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/README.md
+++ b/plugins/backend/catalog-backend-module-okta/README.md
@@ -500,11 +500,11 @@ import { Group } from '@okta/okta-sdk-nodejs';
 function myGroupTransformer(
   group: Group,
   namingStrategy: GroupNamingStrategy,
-  parentGroup: Group | undefined,
   options: {
     annotations: Record<string, string>;
     members: string[];
   },
+  parentGroup: Group | undefined,
 ): GroupEntity {
   // Enrich it with your logic
   const groupEntity: GroupEntity = {


plugins/backend/catalog-backend-module-okta/README.md
```diff
@@ -47,9 +47,9 @@ catalog:
     okta:
       - orgUrl: 'https://tenant.okta.com'
         oauth:
-          clientId: ${OKTA_OAUTH_CLIENT_ID},
-          keyId: ${OKTA_OAUTH_KEY_ID},
-          privateKey: ${OKTA_OAUTH_PRIVATE_KEY},
+          clientId: ${OKTA_OAUTH_CLIENT_ID}
```
punkle, line 50: I know its just a readme.. but do we need a changeset so the readme will be updated in npmjs?

jrtaylorJH: I don't think so?

jrtaylorJH (approved):

plugins/backend/catalog-backend-module-okta/README.md
Xantier: Unfortunately that is needed. NPM central repo doesn't pick up changes on the README file without a release.

Issue #1492, opened 2024-07-23
Security Insights Content Page is not rendering table as expected


david-heidema:
Security Insights Content Page is not rendering table as expected. 

## Current Behavior

When opening the Security Insights tab (EntitySecurityInsightsContent) the table renders stating "No records to display" with the "OPEN" button appearing to be preselected. After selecting the "OPEN" button again, the records display as expected. Clicking "OPEN" a second time also makes the records disappear. The same goes fo the "DISMISSED" button, however, "FIXED" does not have this issue. 
<img width="1370" alt="Screenshot 2024-07-23 at 2 32 18 PM" src="https://github.com/user-attachments/assets/99641a8c-157a-4908-b084-9feaa693f296">

## Expected Behavior

The Security Insights table should initially render with the records displayed that relate to open issues. Likewise the OPEN and DISMISSED buttons should have the capability to be clicked several times in a row without deselecting them(same behavior as FIXED button).

## Steps to Reproduce

Install 2.3.19 of @roadiehq/backstage-plugin-security-insights. View "Security Insights" tab on a valid component.

## Possible Solution

I believe the state of these buttons is being set incorrectly.

## Your Environment

- `@backstage/plugin-catalog-react` version: ^1.21.1
- `@backstage/integration` version: ^1.1.29

david-heidema:
Not stale.

PR #1397, opened 2024-05-22
Show open issues by default in Security Insights & Dependabot

dotboris:
This PR changes the Dependabot & Security Insights tab so that they show open issues by default. This addresses #1395 

image
image

#### :heavy_check_mark: Checklist

- [ ] Added tests for new functionality and regression tests for bug fixes
- [x] Added changeset (run `yarn changeset` in the root)
- [ ] Screenshots of before and after attached (for UI changes)
- [ ] Added or updated documentation (if applicable)

PR #1707, opened 2024-10-30
fix(security-insights): set filtered table data on initial load

pc-bob:
[This PR](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1397) updated this component to filter to `open` on initial load. However, on initial load only `tableData` was being set previously. 

Fixes: https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1492

There is a check in the JSX to render `filteredTableData` if `insightsStatusFilter` is set. Since `insightsStatusFilter` is set with the default state value now (`const [insightsStatusFilter, setInsightsStatusFilter] = useState<SecurityInsightFilterState>('open');`), this was rendering `filteredTableData` on the first render which was null. Clicking on `Open` would display data but may give uses a false sense of security that they have no security insights with a quick glance at the tab. 

  - there are no existing tests for this component
  - Tested this on one of our components that has security insights but don't really want to post screenshots because security
  - N/A - bug fix

### DIFF
--- a/.changeset/fifty-forks-cheat.md
+++ b/.changeset/fifty-forks-cheat.md
@@ -1,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-security-insights': patch
+---
+
+fix bug where security insights tab was empty on initial render until clicking one of the filter buttons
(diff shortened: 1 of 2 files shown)


pc-bob:
This PR is not stale. @punkle could I get some assistance on a review for this one?

pc-bob:
Still looking for some review on this one.

Xantier (approved):

PR #1715, opened 2024-11-11
fix(argocd): update linkUrl logic for multi-instance support

tdabasinskas:
Improved the logic for determining ArgoCD instance URL when multiple instances are configured. The function now correctly looks up instance configuration from `appLocatorMethods` when base URL is not set.

This change ensures proper URL resolution for deployments with multiple ArgoCD instances by:
- Looking up instance config based on instance name
- Falling back to instance `url` if `frontendUrl` is not specified
- Using global base URL as fallback


Xantier (approved):
Thanks!

tdabasinskas:
Thanks for merging this, @Xantier! Can we have it released as per #1716?

Xantier:
> Thanks for merging this, @Xantier! Can we have it released as per https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1716?

Thanks for the contribution. The release pipeline is running now

PR #1747, opened 2024-12-02
feat: upgrading logger

jpdeburgo:
We are making the logger compatible with Backstage's `LoggerService` and winston's `Logger` so that Roadie can be used with newer and older versions of Backstage.

We are doing this by accepting both types of loggers

All the type definitions with a logger have been changed to 
```
import { Logger } from 'winston';
import { LoggerService } from '@backstage/backend-plugin-api';
....
logger: Logger | LoggerService
....
```

Whenever it is required that a logger is from `winston` we do this 
```
import { loggerToWinstonLogger } from '@backstage/backend-common';
import { Logger } from 'winston';

....

const winstonLogger = logger instanceof Logger ? logger : loggerToWinstonLogger(logger);
createRouter({ logger: winstonLogger, config, discovery });

....
```
This way we check to see what type of logger it is and if it is a `LoggerService` type then we will change it to a `winston` `Logger`

### DIFF
--- a/.changeset/two-knives-hammer.md
+++ b/.changeset/two-knives-hammer.md
@@ -1,0 +1,13 @@
+---
+'@roadiehq/scaffolder-backend-module-http-request': minor
+'@roadiehq/scaffolder-backend-argocd': minor
+'@roadiehq/backstage-plugin-argo-cd-backend': minor
+'@roadiehq/rag-ai-backend-embeddings-openai': minor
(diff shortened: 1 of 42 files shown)


Xantier (changes requested):
Thanks, this looks good. Could you add a changeset and we'll get this released?

Xantier (approved):

PR #1754, opened 2024-12-11
 AWS discovery Providers - add support for namespaced group tag values

sblausten:

### DIFF
--- a/plugins/backend/catalog-backend-module-aws/src/utils/tags.ts
+++ b/plugins/backend/catalog-backend-module-aws/src/utils/tags.ts
@@ -108,7 +108,10 @@ export const ownerFromTags = (
 
   if (ownerString && groups && groups.length > 0) {
     const exactMatch = groups.find(
-      g => g.metadata.name.toLowerCase() === ownerString?.toLowerCase(),
+      g =>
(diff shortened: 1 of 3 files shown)


Xantier (approved):

plugins/backend/catalog-backend-module-aws/src/utils/tags.ts
```diff
@@ -108,7 +108,10 @@ export const ownerFromTags = (
 
   if (ownerString && groups && groups.length > 0) {
     const exactMatch = groups.find(
-      g => g.metadata.name.toLowerCase() === ownerString?.toLowerCase(),
+      g =>
+        g.metadata.name.toLowerCase() === ownerString?.toLowerCase() ||
+        `${g.metadata.namespace}/${g.metadata.name}`.toLowerCase() ===
```
punkle, line 113: Do you also need to check for "group:${g.metadata.namespace}/${g.metadata.name}" and "group:${g.metadata.name}"?

sblausten: I think we shouldn't go as far as asking people to add the whole entity ref but could do also

PR #1827, opened 2025-02-19
fix: Remove unnecessary mkdoc techdocs validation from the catalog validator

ColetteDeBeer:
Fixes https://github.com/RoadieHQ/backstage-entity-validator/issues/102

This PR removes the code that validated if a mkdocs file was present in the directory specified for the techdocs annotation. Backstage uses a default mkdocs file if none is found. Please see the issue for more information.

- [x] ~~Added~~ Removed tests for new functionality and regression tests for bug fixes


ColetteDeBeer:
> I am a little on the fence on this one but lean a little on the side of the status quo.. maybe we could make this configurable?

As long as I can use a more recent version of the validator action without it breaking, I'm happy, 😄 but I think making this configurable would be a great middle ground.

PR #1852, opened 2025-03-11
feat(@roadiehq/backstage-plugin-argo-cd-backend) - moved factory inline in ref

fearphage:
Used [`defaultFactory` option](https://backstage.io/docs/reference/backend-plugin-api.servicerefoptions) instead of requiring an external service factory include. 

This reverts parts of #1832. The biggest thing is it **reverts the API break** and allows this module to be loaded in one inline `import()` like all other modules.

I didn't see any tests for this. I verified that the dev server (`yarn start`) started with no issues. I also made this change in the `node_modules` folder when I was trying to upgrade and confirmed that I didn't need an additional include. With the new backend, the `/alpha` include is still all that is needed after this change.


plugins/backend/backstage-plugin-argo-cd-backend/src/refs/argocdService.ref.ts
```diff
@@ -1,6 +1,22 @@
...
+      deps: {
+        config: coreServices.rootConfig,
+        logger: coreServices.logger,
+      },
+      async factory({ config, logger }) {
+        const argoUserName =
+          config.getOptionalString('argocd.username') ?? 'argocdUsername';
+        const argoPassword =
+          config.getOptionalString('argocd.password') ?? 'argocdPassword';
+        return new ArgoService(argoUserName, argoPassword, config, logger);
```
fearphage, line 19: This API is unexpected and seems overly complex. You have access to the entirety of the `config` so it seems odd to pull some pieces of the config out and also pass those along with the entire config. I just copied this from the service factory as is, but I think this could be cleaned up. I'll do it in a follow up pull request.

fearphage:
Tagging @vanessap-aa since they worked on the previous ticket for visibility and insights.

.changeset/soft-lands-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-argo-cd-backend': major
```
Xantier, line 2: Let's make this a minor change since it is not breaking

Xantier (approved):

Issue #1838, opened 2025-02-27
[catalog-backend-module-okta] Update Okta SDK to 7.1.1 to fix


secustor:
To fix https://github.com/advisories/GHSA-3hvj-2783-34x2 in our Backstage installation. 

I have tried to bump `@okta/okta-sdk-nodejs` to [`7.1.1`](https://github.com/okta/okta-sdk-nodejs/releases/tag/okta-sdk-nodejs-7.1.1), though this require extensive changes to make it testable and some refactoring. 

Before going through with it I wanted to check if such PRs would be even accepted. 

Thx!

faizhasim:
We reach out to roadie security a while back. Do let community know how can we help.

Xantier:
Thanks @secustor , PRs are most definitely welcome. Let me know when you have the fix ready and I can merge it in for you

helzahalim:
Let us know how we can help @secustor

secustor:
K, can you assign the issue to me? 

I will start to work on this soonish™

PR #1854, opened 2025-03-12
chore: make repo buildable node 22

secustor:
For a better review experience review the commits should be reviewed separately. 

- bumped to `better-sqlite3` and `dockerode` to Node 22 compatible versions
- Run `yarn install` which cleaned up the `package.json` and reordered the entries alphabetically. 
-  ~~Added volta configuration so it does not break the setup for users of it.~~ 

Context: started working on https://github.com/RoadieHQ/roadie-backstage-plugins/issues/1838

I recommend checking why Dependabot is not picking up these dependencies. 
Additionally a lot of Github actions are behind major versions 


package.json
```diff
@@ -56,5 +56,9 @@
     "typescript": "^5.5.4"
   },
   "prettier": "@spotify/prettier-config",
+  "volta": {
```
planeiii-te, line 59: I cannot merge this, but I think it would be a lot easier to merge if it was only doing what it said in the title. Perhaps add this tool/feature/functionality in another PR? Just sharing my thoughts.

secustor: I actually ran a lot of problems because of mismatched versions of different tools like yarn, node and python.

The idea is to elevate this to some degree. If that is the only problem preventing a merge, I happily remove it.

Alternatively there should be something like `.tools-version` which documents the actually used versions so tooling can pick it up.

Xantier: Thanks for the explanation secustor. At this point it's unlikely that we are tying ourselves into a random tool manager.

PR #1872, opened 2025-03-21
feat(@roadiehq/backstage-plugin-argo-cd-backend): added functionality that would allow us to login using azure active directory

vanessap-aa:
Added functionality that would allow us to login using azure active directory. This is non-breaking.


plugins/backend/backstage-plugin-argo-cd-backend/README.md
```diff
@@ -89,6 +89,46 @@ argocd:
...
+    name: Azure
+    issuer: https://login.microsoftonline.com/<ClientId>/v2.0
+    clientID: <ClientId>
+    clientSecret: $oidc.azure.clientSecret
+    requestedIDTokenClaims:
+      groups:
+          essential: true
+    allowedAudiences:
+      - <ClientId>
+<ClientIdOfClientCredSp> # If a different service principal is being used for client_credentials authentication
```
Jay-Madden, line 129:
```suggestion
     - <ClientIdOfClientCredSp> # If a different service principal is being used for client_credentials authentication
```

plugins/backend/backstage-plugin-argo-cd-backend/README.md
```diff
@@ -89,6 +89,46 @@ argocd:
...
+    name: Azure
+    issuer: https://login.microsoftonline.com/<ClientId>/v2.0
+    clientID: <ClientId>
+    clientSecret: $oidc.azure.clientSecret
+    requestedIDTokenClaims:
+      groups:
+          essential: true
+    allowedAudiences:
+      - <ClientId>
+     - <ClientIdOfClientCredSp> # If a different service principal is being used for client_credentials authentication
```
vanessap-aa, line 129:
```suggestion
    allowedAudiences:
      - <ClientId>
      - <ClientIdOfClientCredSp> # If a different service principal is being used for client_credentials authentication
```

plugins/backend/backstage-plugin-argo-cd-backend/package.json
```diff
@@ -60,7 +60,8 @@
     "cross-fetch": "^3.1.4",
     "express": "^4.17.1",
     "express-promise-router": "^4.1.0",
-    "winston": "^3.2.1"
+    "winston": "^3.2.1",
+    "qs": "^6.10.3"
```
fearphage, line 64: I would recommend against editing `package.json` files by hand.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -32,8 +32,10 @@ import {
   GetArgoApplicationFetchResponse,
   getArgoApplicationInfoProps,
   terminateArgoAppOperationProps,
+  AzureConfig,
```
fearphage, line 35: Why is this named Azure config instead of something more generic like OIDC config?

.changeset/tasty-turkeys-brake.md
```diff
@@ -0,0 +1,61 @@
...
+You can now use azure active directory (Microsoft Entra ID) to authentication to Argo CD instances!
+
+**Login Priorities** (We will prioritize logins as follows):
+
+1. Instance Level Tokens
+
+   ```json
+       argocd:
+           appLocatorMethods:
+               - type: 'config'
```
fearphage, line 16: There's no need to add quotes around `config`. It's a string just like all the other values.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -207,32 +209,64 @@ export class ArgoService implements ArgoServiceApi {
...
     }
-    if (resp.status === 401) {
-      throw new Error(`Getting unauthorized for Argo CD instance ${url}`);
+
+    if (azureCredentials) {
+      const resp = await fetch(
+        `${azureCredentials.loginUrl}/${azureCredentials.tenantId}/oauth2/v2.0/token`,
+        {
+          method: 'POST',
+          body: qs.stringify({
```
fearphage, line 246: I'm not familiar with `qs`. If you just need query strings encoded, couldn't you just use `URLSearchParams`? Then you wouldn't need the library at all.

```suggestion
          body: new URLSearchParams({
```

.changeset/tasty-turkeys-brake.md
```diff
@@ -0,0 +1,61 @@
...
+              - type: 'config'
+              instances:
+                  - name: argoInstance1
+                    url: https://argoInstance1.com
+   ```
+
+4. Azure Credentials using azure login url
+
+   ```json
+      azure:
```
fearphage, line 54: There's already a `type` property on the existing configs. Could that be used to bring this inline with that config? Instead of `type: config` it could be `type: oidc` or something of that nature.

fearphage (changes requested):

.changeset/tasty-turkeys-brake.md
vanessap-aa: Hey @fearphage, It looks like we don't have much logic using this `type` field if its ANY other value. The code only looks at the argo instances if they have a type 'config'. That said, we can add more logic to support different "types", however, I feel that doing so would introduce more changes than necessary at the moment. As it stands, I do not think the `type` field is providing any value.

.changeset/tasty-turkeys-brake.md
vanessap-aa, line 16:
```suggestion
               - type: config
```

.changeset/tasty-turkeys-brake.md
```diff
@@ -0,0 +1,61 @@
...
+                   url: https://argoInstance1.com
+                   token: ${ARGOCD_AUTH_TOKEN} # Instance Level Token
+   ```
+
+2. Instance Level Username and Password
+
+   ```json
+      argocd:
+          appLocatorMethods:
+              - type: 'config'
```
vanessap-aa, line 28:
```suggestion
              - type: config
```

.changeset/tasty-turkeys-brake.md
```diff
@@ -0,0 +1,61 @@
...
+
+3. Upper Level Username and Password
+
+   ```json
+      argocd:
+          # Upper Level Username and Password
+          username: ${ARGOCD_USERNAME}
+          password: ${ARGOCD_PASSWORD}
+          appLocatorMethods:
+              - type: 'config'
```
vanessap-aa, line 45:
```suggestion
              - type: config
```

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
vanessap-aa: Hey! Thanks for the suggestion. We learned something new. We went ahead and changed it and we are removing the package.

plugins/backend/backstage-plugin-argo-cd-backend/package.json
vanessap-aa: We installed this via yarn. The yarn lock didn't update because its existing.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
vanessap-aa: Its not more generic because its specific to azure login. What we will do is change the code to add the flexibility of adding other oidc login configs for later contributions.

vanessap-aa: Now looking at this, what you are pointing to is a type. So with respect to the type, i think this is accurate. However, we will still make the change I previously mentioned.

vanessap-aa:
@fearphage We completed what we felt was needed to change. Please re-review at your earliest convenience.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
fearphage: I was mostly talking about the naming convention. Nothing else in this file is named for RackSpace, AWS, GCP, etc. They are named based on the tech instead of the service provider. I was wondering why we were breaking that convention now.

plugins/backend/backstage-plugin-argo-cd-backend/package.json
fearphage: I've never seen `yarn` install a package out of order. Are you on a very old version of `yarn`?


```shell
➜  roadie-backstage-plugins git:(main) ✗ yarn --version
1.22.22
```

fearphage (approved):
The changes

Xantier (approved):

PR #1878, opened 2025-03-25
chore: upgraded `better-sqlite3` to support node `v22`

fearphage:
Upgraded `better-sqlite` so it was usable with Node `v22`. I couldn't install in Node 22 without this change.

This is similar to #1854, but I ran exactly one command:

```shell
yarn --cwd packages/backend add better-sqlite3@latest
```

> [!IMPORTANT]
>
> You're supporting version 18 of Node which isn't supported in Backstage anymore.


packages/backend/package.json
```diff
@@ -66,11 +66,11 @@
     "winston": "^3.2.1"
   },
   "devDependencies": {
-    "@types/luxon": "^3.0.0",
```
fearphage, line 69: It might be worthwhile to add a GitHub action to check for people hand editing `package.json` files.

Xantier (approved):

PR #1887, opened 2025-03-31
show reasonable error if http request fails

punkle:
show reasonable error if http request fails

### DIFF
--- a/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
+++ b/plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
@@ -114,6 +114,17 @@ const SelectFieldFromApiComponent = (
       })}?${params}`,
       { headers },
     );
+    if (options.successStatusCode) {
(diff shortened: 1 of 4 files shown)


plugins/scaffolder-field-extensions/scaffolder-frontend-module-http-request-field/src/components/SelectFieldFromApi/SelectFieldFromApi.tsx
```diff
@@ -114,6 +114,17 @@ const SelectFieldFromApiComponent = (
       })}?${params}`,
       { headers },
     );
+    if (options.successStatusCode) {
+      if (response.status !== options.successStatusCode) {
+        throw new Error(
+          `Failed to retrieve data from API: ${response.statusText}`,
+        );
+      }
+    } else if (!response.ok) {
```
punkle, line 123: Maybe by default it should default to the existing behavior and we can make this minor, thoughts?

punkle:
@Xantier the issue here is that where an error is returned there is no error from the http client.. the error occurs when the data field attempts to retrieve the dropdown elements, labels/values etc from the error response body.

I feel we should allow the status code to be checked to give an error before it parses the body.

Xantier (approved):

PR #1889, opened 2025-03-31
fix(catalog-backend-module-okta): chunk membership resolution queries to avoid socket overloading

alecjacobs5401:
Recently, we expanded our usage of the `OktaGroupEntityProvider` to pull in more Groups than we typically had in the past (>1000).

We started noticing that every run of the provider would produce different results and that Groups were disappearing on subsequent runs. After some local investigation and debugging, we found that many of the `listUsers()` queries were silently failing and thus skipping that group entirely for ingestion.

For us, the underlying root cause was `ERR_SOCKET_TIMEOUT` since we were kicking off >1000 async requests to Okta to resolve membership. This is resolved by adding chunking to membership resolution to limit how many async processes we have in flight at once.

Implementing this change and testing it locally showed minimum to no difference in timing to resolve all membership for 1000+ groups, likely due to lack of socket contention meaning things were resolving faster and/or automated rate limit handling happening internally to the Okta package not needing to take effect.

This is an extension of https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1280 which was only applied to `OktaOrgEntityProvider`

We separately schedule the User and Group providers to avoid hammering Okta and to reduce risk.

- [x] Screenshots of before and after attached (for UI changes): **N/A**

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
+++ b/plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import chunk from 'lodash/chunk';
 import { GroupEntity } from '@backstage/catalog-model';
 import { Config } from '@backstage/config';
 import { OktaEntityProvider } from './OktaEntityProvider';
@@ -36,6 +37,8 @@ import { getParentGroup } from './getParentGroup';
 import { OktaGroupEntityTransformer } from './types';
 import { LoggerService } from '@backstage/backend-plugin-api';
 
+const DEFAULT_CHUNK_SIZE = 250;
+
 /**
  * Provides entities from Okta Group service.
  */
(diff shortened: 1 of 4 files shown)


alecjacobs5401:
FYI @Xantier @punkle

plugins/backend/catalog-backend-module-okta/src/providers/OktaGroupEntityProvider.ts
```diff
@@ -129,60 +136,73 @@ export class OktaGroupEntityProvider extends OktaEntityProvider {
       logger: this.logger,
     });
 
-    await Promise.allSettled(
-      Object.entries(oktaGroups).map(async ([_, group]) => {
-        const members: string[] = [];
-        await group.listUsers().each(user => {
+    for (const entries of chunk(Object.entries(oktaGroups), this.chunkSize)) {
```
punkle, line 139: I feel like p-map might be a good choice for this type of thing.. but Im not sure if it has the allSettled type behaviour.

punkle (approved):

PR #1925, opened 2025-06-11
Fix ArgoCD plugin throwing errors in NFS

benjdlambert:
The `argocd` plugin uses the wrong Blueprint in the new frontend system throwing errors because the `argoCdPage` is actually an `EntityContent` tab instead and expects an `EntityContext` to be available.

This fixes that by moving this to be an `EntityContentBlueprint` instead.


punkle:
Thank you 🙏

benjdlambert:
Thank you! :pray:

PR #1928, opened 2025-06-11
fix(wiz): corrected url in frontend plugin to the correct backend plugin

jneate:
This change allows wiz plugin consumers to add the backend plugin using the new Backstage plugin system, and removes the need for additional customisation or additional development effort.

It could be achieved by changing the backend component's plugin ID but I wasn't sure if there were other instances referencing this plugin, hence the simpler change to the frontend plugin to change the URL being called by the API client.

Existing tests pass, mainly a documentation change.

Marked the change as breaking because if people have customised the backend plugin, their integration won't work with the new frontend plugin anymore.

### DIFF
--- a/.changeset/old-teachers-sniff.md
+++ b/.changeset/old-teachers-sniff.md
@@ -1,0 +1,35 @@
+---
+'@roadiehq/plugin-wiz-backend': major
+---
+
+**BREAKING** Fixed an issue with backend plugin initialisation approach which required additional configuration for the end user.
+
+This fix allows plugin consumers to add the backend plugin using the new backend system, unfortunately, this breaks existing approaches as plugin consumers will have customised the backend plugin to respond on the incorrect URL.
+
+These changes are **required** to `packages/backend/src/index.ts`
+
+```diff
++ // Install wiz backend plugin
++ backend.add(import('@roadiehq/plugin-wiz-backend'));
+- import wiz from './plugins/wiz';
+- // ...
+- async function main() {
+-   // ...
+-   const wizEnv = useHotMemoize(module, () => createEnv('wiz'));
+-
+-   const wizConfig = {
+-   clientId: config.getOptionalString('wiz.clientId'),
+-   clientSecret: config.getOptionalString('wiz.clientSecret'),
+-   tokenUrl: config.getOptionalString('wiz.tokenUrl'),
+-   apiUrl: config.getOptionalString('wiz.wizAPIUrl'),
+- };
+-
+-   const apiRouter = Router();
+-     if (wizConfig.enabled && wizConfig.clientId && wizConfig.clientSecret && wizConfig.tokenUrl && wizConfig.apiUrl) {
+-     router.use('/wiz-backend', await wiz(wizEnv));
+-   } await wiz(wizEnv));
+-   // ...
+- }
+```
+
+And you can remove the previous plugin customisation file `packages/backend/src/plugins/wiz.ts`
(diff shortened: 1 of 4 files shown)


.changeset/polite-dogs-rescue.md
```diff
@@ -0,0 +1,36 @@
+---
+'@roadiehq/backstage-plugin-wiz': major
```
punkle: Could we make this a minor by adding a config option to the config.d.ts file with frontend visibility and defaulting it to `wiz-backend`?

jneate: we could, but it feels overkill adding a config option to support the expected behaviour of the backend plugin system

I'm happy to make the change if you'd rather do that though and avoid the breaking change?

punkle: I think as it currently stands the backend must be mounted at `wiz-backend`, after this frontend release, the backend will need to be either moved to backend-wiz-backend or be mounted in both places if it is in use by other parts of the application? Isn't that just more cumbersome forcing everyone to change the mounting to take the update? But if we add flexibility it might be easier to migrate?

I am not familiar with backend plugins using this type of name, generally it would be just "wiz" right? So Im unsure why we are prefixing it with the word plugin- in the first place? Could you explain the motivation there?

jneate: The existing backend plugin has the prefix `plugin-wiz-backend` and I'm in agreement with you I don't know why it was created like that, the other option as mentioned in the PR is to change the plugin ID to `wiz-backend` and drop the `plugin-` prefix which was the other approach I was thinking about, but that might still be seen as a breaking change because I don't know what else would, or could, reference the plugin ID

If the ID was changed then there'd be no changes required to the frontend plugin, it'd be self contained to the backend one. I'd rather remove the prefix from the backend plugin than add config into the frontend plugin but again it depends whether you'd want to mark that change as a major or breaking change and the overhead that comes with?

jneate: @punkle I've updated the PR based on this chat, I still think it's worth marking the backend plugin as a major/breaking change to ensure people are aware of the changes but happy to take your steer

jneate:
@Irma12 happy to discuss/compromise if you disagree with the approach as per the comments with @punkle

kissmikijr:
I'm a bit confused that you mention multiple times frontend but all of the changes are in the backend plugin ? 🤔

jneate:
@kissmikijr the first version changed the frontend plugin to use the correct backend URL, following feedback from others on the PR it was changed to keep the frontend as it was, and change the backend plugin which would alter the default URL meaning no changes needed in the frontend

kissmikijr:
Yea, fair, do you mind changing the descr and title please? I'm in favor of this change. I think it's fine to change the pluginId and leave it as a major version change. 

However the best would be if we would update the pluginId to be only wiz and change the frontend also to expect that :D

jneate:
@kissmikijr are you happy for the frontend change to be a breaking one also?
People upgrading to the newer version would have to upgrade to the newer backend version also so it makes sense in my head

If you're happy with the above I'll update the PR today

kissmikijr:
@jneate Yes, we are fine with a breaking change in the FE and BE in favor of the rename to use only `wiz` as the plugin id. Should be cleaner overall and the "migration" path will be super simple as well!

I would really appreciate if you would do it!

jneate:
@kissmikijr @Irma12 - ready for review following the above changes / discussion

kissmikijr (approved):
🚀 Thank you!

PR #1930, opened 2025-06-11
remove more of the backend common package

punkle:
remove more of the backend common package

### DIFF
--- a/.changeset/chatty-guests-sing.md
+++ b/.changeset/chatty-guests-sing.md
@@ -1,0 +1,7 @@
+---
+'@roadiehq/backstage-plugin-argo-cd-backend': patch
+'@roadiehq/catalog-backend-module-okta': patch
+'@roadiehq/backstage-plugin-aws-backend': patch
+---
+
+Remove deprecated `@backstage/backend-common` from dependencies.


Xantier (approved):

plugins/backend/backstage-aws-backend/src/service/router.test.ts
```diff
@@ -18,13 +18,13 @@ import { createRouter } from './router';
 import express, { Express } from 'express';
 import { ConfigReader } from '@backstage/config';
 import request from 'supertest';
-import { getVoidLogger } from '@backstage/backend-common';
 import { CloudControl, GetResourceCommand } from '@aws-sdk/client-cloudcontrol';
 import { mockClient } from 'aws-sdk-client-mock';
+import { createLogger } from 'winston';
```
Xantier, line 23: test-utils should have something for this so you don't need to construct an actual one. Doesn't matter really but would make cleaner test logs

punkle: I did try that, but I couldnt get that to work straight away so I shevled the idea for expediency.. but we can revisit.

PR #1937, opened 2025-06-24
Add validator overrides

sblausten:
Make entity validation rules consistent with Roadie rules by loosening some restrictions. 

- Tags can be any string less that 64 chars
- Label values can be any string
- names can be up to 120 chars 


utils/roadie-backstage-entity-validator/src/validator.js
```diff
@@ -73,6 +74,26 @@ export const validate = async (
   customAnnotationSchemaLocation = '',
 ) => {
   let validator;
+
+  const overrides = {
```
punkle, line 78: We probably could add tests for this?

sblausten: We are loosening restrictions so I don't think there's a need here

PR #1939, opened 2025-06-25
Release script fix

sblausten:

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,10 +42,12 @@
   "devDependencies": {
     "@backstage/cli": "^0.29.2",
     "@changesets/cli": "^2.18.0",
+    "@octokit/rest": "^22.0.0",
     "@spotify/prettier-config": "^14.0.0",
     "@types/webpack": "^5.28.0",
     "concurrently": "^7.0.0",
     "eslint": "^8.6.0",
+    "eslint-plugin-import": "^2.32.0",
     "eslint-plugin-notice": "^0.9.10",
     "eslint-plugin-react": "^7.28.0",
     "eslint-plugin-unused-imports": "^3.1.0",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -42,10 +42,12 @@
   "devDependencies": {
     "@backstage/cli": "^0.29.2",
     "@changesets/cli": "^2.18.0",
+    "@octokit/rest": "^22.0.0",
```
punkle, line 45: ?

punkle:
Could you explain this change a little bit. Im must admit, I am a bit lost.

package.json
sblausten: We have a release script that uses this dependancy. It isn't listed here though. You can see the script in the PR

sblausten:
> Could you explain this change a little bit. Im must admit, I am a bit lost.

It fixes this issue which happens everytime I bump versions and expect a release. https://github.com/RoadieHQ/roadie-backstage-plugins/actions/runs/15871773070/job/44750222533#step:10:12

punkle:
> It fixes this issue which happens everytime I bump versions and expect a release

How come the PR also includes changes to the validator? Why do you think its failing now? It seems that release process has been in place for 3/4 years.

sblausten:
Maybe I'm doing something wrong - we raise automated prs with a title `Version Packages <pr-number>` which doesn't match the script regex so the release step always will fail. To make it pass the automated prs need to be renamed currently. It could be that this is how it is supposed to work but I don't see a GitHub release for my latest change even after the new version is available on npm which seems wrong

punkle:
The latest version is 2.6.0 right? It seems to be here: https://www.npmjs.com/package/@roadiehq/roadie-backstage-entity-validator/v/2.6.0

PR #1964, opened 2025-07-15
add ability to template aws entities

punkle:
Add ability to template AWS provided entities

The idea is that the entity provider can be configured with a template. The template would be rendered with the eks cluster returned from the aws apis. This allows more flexibility as to the resulting entity that is produced by the entity provider.

e.g. the provider could be configured like this:

```javascript
  const tempate = `kind: Resource
apiVersion: backstage.io/v1beta1
metadata:
  name: {{ cluster.name | to_entity_name }}
  annotations:
    amazon.com/eks-cluster-arn: {{ cluster.arn }}
    amazon.com/iam-role-arn: {{ cluster.roleArn }}
    kubernetes.io/auth-provider: aws
    kubernetes.io/x-k8s-aws-id: {{ cluster.name }}
  title: {{ accountId }}:{{ region }}:{{ cluster.name }}
  labels:
    some_url: {{ cluster.tags.some_url | replace(":", "-") | replace("/", "-") }}`;

  const eksClusterProvider = AWSEKSClusterProvider.fromConfig(
    config,
    {
      logger,
      template
    },
  );
```

### DIFF
--- a/plugins/backend/catalog-backend-module-aws/src/providers/AWSEKSClusterProvider.ts
+++ b/plugins/backend/catalog-backend-module-aws/src/providers/AWSEKSClusterProvider.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { ResourceEntity } from '@backstage/catalog-model';
+import { Entity } from '@backstage/catalog-model';
 import { EKS, paginateListClusters } from '@aws-sdk/client-eks';
 import type { Logger } from 'winston';
 import { LoggerService } from '@backstage/backend-plugin-api';
@@ -40,17 +40,22 @@ import {
 import { CatalogApi } from '@backstage/catalog-client';
 import { AccountConfig, DynamicAccountConfig } from '../types';
 import { duration } from '../utils/timer';
+import { compile, Template, Environment } from 'nunjucks';
+import yaml from 'js-yaml';
+import crypto from 'crypto';
 
 /**
  * Provides entities from AWS EKS Cluster service.
  */
 export class AWSEKSClusterProvider extends AWSEntityProvider {
   private readonly clusterTypeValue: string;
+  private readonly template: Template | undefined;
 
   static fromConfig(
     config: Config,
     options: {
       logger: Logger | LoggerService;
+      template?: string;
       catalogApi?: CatalogApi;
       providerId?: string;
       ownerTag?: string;
@@ -75,6 +80,7 @@ export class AWSEKSClusterProvider extends AWSEntityProvider {
     account: AccountConfig,
     options: {
       logger: Logger | LoggerService;
+      template?: string;
       catalogApi?: CatalogApi;
       providerId?: string;
       ownerTag?: string;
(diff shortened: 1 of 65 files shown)


kissmikijr:
What does it mean to template eks cluster entity?

punkle:
> What does it mean to template eks cluster entity?

The idea is that the entity provider can be configured with a template. The template would be rendered with the eks cluster returned from the aws apis. This allows more flexibility as to the resulting entity that is produced by the entity provider.

e.g. the provider could be configured like this:

```javascript
  const tempate = `kind: Resource
apiVersion: backstage.io/v1beta1
metadata:
  name: {{ cluster.name | to_entity_name }}
  annotations:
    amazon.com/eks-cluster-arn: {{ cluster.arn }}
    amazon.com/iam-role-arn: {{ cluster.roleArn }}
    kubernetes.io/auth-provider: aws
    kubernetes.io/x-k8s-aws-id: {{ cluster.name }}
  title: {{ accountId }}:{{ region }}:{{ cluster.name }}
  labels:
    some_url: {{ cluster.tags.some_url | replace(":", "-") | replace("/", "-") }}`;

  const eksClusterProvider = AWSEKSClusterProvider.fromConfig(
    config,
    {
      logger,
      template
    },
  );
```

plugins/backend/catalog-backend-module-aws/src/providers/AWSEKSClusterProvider.ts
```diff
@@ -166,42 +183,56 @@ export class AWSEKSClusterProvider extends AWSEntityProvider {
...
+          let entity: Entity;
+          if (this.template) {
+            entity = yaml.load(
+              this.template.render({
+                cluster: cluster.cluster,
+                accountId,
+                region: this.region,
+              }),
+            ) as Entity;
+            console.log(entity);
```
Xantier:
```suggestion
```

plugins/backend/catalog-backend-module-aws/src/providers/AWSEntityProvider.ts
```diff
@@ -71,6 +77,41 @@ export abstract class AWSEntityProvider implements EntityProvider {
...
+      });
+      env.addFilter('split', function (str, delimiter) {
+        return str.split(delimiter);
+      });
+      this.template = compile(options.template, env);
+    }
+  }
+
+  protected renderEntity(
+    context: any,
```
Xantier: This needs some typing. You are expanding it like an object or array on line 103 so it can't be any

punkle:
> Either write docs on each provider that tells at least the name of the object that is available to the templating functionality or rename it to something generic

Done

plugins/backend/catalog-backend-module-aws/src/providers/AWSEntityProvider.ts
punkle: Done.

Xantier (approved):

PR #1995, opened 2025-07-24
add default annotions to the cluster resource

punkle:
add default annotions to the cluster resource


plugins/backend/catalog-backend-module-aws/src/providers/AWSEKSClusterProvider.ts
```diff
@@ -170,9 +170,12 @@ export class AWSEKSClusterProvider extends AWSEntityProvider {
...
 
           const labels = this.labelsFromTags(cluster.cluster?.tags);
-          let entity: Entity | undefined = this.renderEntity({
-            data: cluster.cluster,
-          });
+          let entity: Entity | undefined = this.renderEntity(
+            {
+              data: cluster.cluster,
+            },
+            { defaultAnnotations: await defaultAnnotations },
```
punkle, line 177: This is the main change. It was missed in the initial pr.

kissmikijr, line 177: It looks unique that we await something that is not callable ? 🤔 why don't we await the default annotations at assingment?

kissmikijr (changes requested):
question

plugins/backend/catalog-backend-module-aws/src/providers/AWSEKSClusterProvider.ts
punkle: Yeah I dont know. I didnt change that.

punkle: I guess its slightly more optimal, because it can list the lambda functions while its waiting for the annotations to be retrieved.

kissmikijr (approved):

Issue #2004, opened 2025-07-30
@roadiehq/roadie-backstage-entity-validator 2.7.0 misses dist folder


timo-reymann:
## Expected Behavior

The dist code is present and the package is useable

## Current Behavior

It fails due to the missing dist folder:

```
[...]@roadiehq/roadie-backstage-entity-validator/dist/index.js'. Please verify that the package.json has a valid "main" entry
```

## Steps to Reproduce

- Install @roadiehq/roadie-backstage-entity-validator in v2.7.0 
- Try to use the library code

## Possible Solution

- Release hotfix

## Context

- Librar is unusable

## Your Environment

- `@roadiehq/roadie-backstage-entity-validator` version: 2.7.0

timo-reymann:
Also see the npm registry code tab: https://www.npmjs.com/package/@roadiehq/roadie-backstage-entity-validator/v/2.7.0?activeTab=code

LeonardDrs:
Hi @kissmikijr 

Looks like this is due to the [transition](https://github.com/RoadieHQ/roadie-backstage-plugins/pull/1982/files) from `lerna` to `yarn` to publish packages. This would have been fine if `yarn@1.22.22` was still used - as it [used to call](https://github.com/yarnpkg/yarn/blob/master/src/cli/commands/publish.js#L47) the `prepare` script back in the day, but  now it [doesn't](https://github.com/yarnpkg/berry/blob/%40yarnpkg/cli/4.9.2/packages/plugin-npm-cli/sources/commands/npm/publish.ts#L97).

We can see indeed that the [release job](https://github.com/RoadieHQ/roadie-backstage-plugins/actions/runs/16467873142/job/46549812151) didn't call for the `prepare` lifecycle who's in charge of [building](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/utils/roadie-backstage-entity-validator/package.json#L26) the @roadiehq/roadie-backstage-entity-validator module.

Also as a bonus, could we add a `files` property on the package.json file?
It's done on the [other packages](https://github.com/RoadieHQ/roadie-backstage-plugins/blob/main/plugins/backend/backstage-aws-backend/package.json#L56-L59) to avoid publish the source files in the npm package.

Thanks a lot for the time spent on this issue!

kissmikijr:
Ahh, nice thank you, yea we will get this sorted. We had some rogue packages which didnt get the proper setup after the migration!

PR #2009, opened 2025-08-07
move validator into the backstage plugins

kissmikijr:
Fixes: #2004


plugins/roadie-backstage-entity-validator/src/relativeSpaceValidation.ts
```diff
@@ -63,6 +63,6 @@ export const relativeSpaceValidation = async (
       }),
     );
   } catch (e) {
```
punkle, line 65:
```suggestion
  } catch {
```

plugins/roadie-backstage-entity-validator/src/relativeSpaceValidation.ts
```diff
@@ -48,12 +48,12 @@ const validateTechDocs = async (data, filePath) => {
...
-  fileContents,
-  filePath,
-  verbose,
+  fileContents: string,
+  filePath: string,
+  verbose: boolean,
 ) => {
   try {
-    const data = yaml.loadAll(fileContents, { schema: yaml.CORE_SCHEMA });
+    const data = yaml.loadAll(fileContents, null, { schema: yaml.CORE_SCHEMA });
```
punkle, line 56: This seems like a material change? Or maybe a new version of the yaml lib requires this?

yarn.lock
```diff
@@ -15475,9 +15475,9 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@roadiehq/roadie-backstage-entity-validator@workspace:utils/roadie-backstage-entity-validator":
+"@roadiehq/roadie-backstage-entity-validator@workspace:plugins/roadie-backstage-entity-validator":
```
punkle, line 15478: 😍

punkle (approved):

plugins/roadie-backstage-entity-validator/src/relativeSpaceValidation.ts
kissmikijr: I think loadAll expected the options in the 3rd positional paramter but the file was regular js so TS didnt check and now that I converted it to TS it complains that it should be 3rd 🤔 So I think it's a fix actually

punkle: ok.. 👍  Hopefully the tests will catch it if its broken.

plugins/roadie-backstage-entity-validator/src/validator.ts
```diff
@@ -50,7 +51,7 @@ const VALIDATORS = {
   product: ajvCompiledJsonSchemaValidator(productSchema),
 };
 
-function modifyPlaceholders(obj) {
+function modifyPlaceholders(obj: any) {
```
punkle, line 54: I know this isnt your code. Maybe we should do a pass on the syntax here to use more modern syntax.

plugins/roadie-backstage-entity-validator/src/validator.ts
```diff
@@ -16,21 +16,22 @@ import {
...
 
 const ajv = new Ajv({ verbose: true });
 ajvFormats(ajv);
 
-function ajvCompiledJsonSchemaValidator(schema) {
+function ajvCompiledJsonSchemaValidator(schema: any) {
   return {
-    async check(data) {
+    async check(data: any) {
       return entityKindSchemaValidator(schema)(data) === data;
```
punkle, line 35: Not your issue... Im thinking out loud.

We are effectively checking the schemas are schemas at runtime here.. if they arent schemas im assuming it will fail here. maybe it makes sense to convert those json import files to ts with type checking..

plugins/roadie-backstage-entity-validator/src/validator.ts
kissmikijr: Feel free to do it when you want to. This PRs aim is to fix the missing version... for this i needed to make this JS file into TS file...

punkle:
Typescript is identifying loads of areas of potential improvement.

Id vote for getting this pr in and then putting the refactor on the long finger.

punkle (approved):

plugins/roadie-backstage-entity-validator/src/validator.ts
punkle: There is a lot of stuff that typescript is picking up here.. If we do address it, we should do it after this change is merged. Im happy to do a pass if I can get a minute or two.

PR #2013, opened 2025-08-15
fix: allow OIDC auth by not defaulting ArgoCD username/password

snehagupta-coder:
## Description
Fix ArgoCD backend plugin to properly handle OIDC authentication by removing default username/password values.

### Problem
The plugin always provided default username/password even when OIDC config was present, causing authentication to fail when using OIDC.

### Solution
- Remove default values for username/password in ArgocdService constructor
- Only pass username/password if they are explicitly configured
- Add test case for OIDC-only authentication scenario

### Testing
- Added new test case to verify OIDC-only configuration works
- Existing tests pass without modification

### Related Issue
Fixes authentication failures when using OIDC with ArgoCD


snehagupta-coder:
Hello, any update on this issue?

PR #2025, opened 2025-08-27
Use first source if multiple are used

ctria:
ArgoCD applications can have either "source" or multiple "sources" this patch allows the creation of application history for apps with multiple sources by using the first defined source.


Xantier (approved):
Thanks. I was thinking about asking to make this a patch change, but it is technically minor so let's merge it in as is.

PR #2026, opened 2025-08-27
add relative time and display date on hover

pshenbagadelip:
we should use the 2024-08-23 format to be a global standard instead of using a non standard format for the assigned date column


pshenbagadelip:
@Xantier added a small change in the date representation.

Xantier:
Currently the date format is displayed as the user wants it, based on their browser settings. I think we should keep it as is

pshenbagadelip:
Before:
image

After:
my-jira-tickets-card

pshenbagadelip:
> Currently the date format is displayed as the user wants it, based on their browser settings. I think we should keep it as is

After discussing with the team, we felt that displaying relative time (e.g., “5 mins ago”, “2 days ago”) instead of the actual date makes the interface more intuitive. To preserve full context, the exact assigned date can be shown on hover.

For ticket ordering, it makes the most sense to sort primarily by the Assigned column, so that newly assigned tickets naturally float to the top. This keeps the dashboard focused on the most recent activity while maintaining clarity and responsiveness.

Added the screenshots for the same

PR #2036, opened 2025-09-05
fix: argocd backend method not allowed error due to malformed url (trailing slash)

evanshortiss:
The URL join behaviour in the Argo CD plugin was causing "Method Not Allowed" errors if the configured instance URL contains a trailing slash. Understandably the fix is simply not to include a trailing slash, but this PR might help some folks.

### DIFF
--- a/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
+++ b/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
@@ -103,43 +103,43 @@ describe('ArgoCD service', () => {
   });
 
   describe('getRevisionData', () => {
-    it.each([
-      'https://argoinstance1.com',
-      'https://argoinstance1.com/'
-    ])('should get revision data and handle trailing slashes in urls', async (url) => {
-      const expectedResp = {
-        author: 'testuser',
-        date: '2023-03-20T18:44:10Z',
-        message: 'Update README.md',
-      };
+    it.each(['https://argoinstance1.com', 'https://argoinstance1.com/'])(
+      'should get revision data and handle trailing slashes in urls',
+      async url => {
+        const expectedResp = {
+          author: 'testuser',
+          date: '2023-03-20T18:44:10Z',
+          message: 'Update README.md',
+        };
 
-      fetchMock.mockResponseOnce(JSON.stringify(expectedResp));
+        fetchMock.mockResponseOnce(JSON.stringify(expectedResp));
 
-      const name = 'testApp';
-      const token = 'testToken';
-      const revisionId = '15db63ac922a920f388bd841912838ae4d126317';
+        const name = 'testApp';
+        const token = 'testToken';
+        const revisionId = '15db63ac922a920f388bd841912838ae4d126317';
 
-      const resp = await argoService.getRevisionData(
-        url,
-        { name },
-        token,
-        revisionId,
-      );
+        const resp = await argoService.getRevisionData(
+          url,
+          { name },
+          token,
+          revisionId,
+        );
 
-      expect(fetchMock).toHaveBeenCalledWith(
-        expect.stringMatching(
-          `https://argoinstance1.com/api/v1/applications/${name}/revisions/${revisionId}/metadata`,
-        ),
-        expect.objectContaining({
-          headers: {
-            'Content-Type': 'application/json',
-            Authorization: `Bearer ${token}`,
-          },
-        }),
-      );
+        expect(fetchMock).toHaveBeenCalledWith(
+          expect.stringMatching(
+            `https://argoinstance1.com/api/v1/applications/${name}/revisions/${revisionId}/metadata`,
+          ),
+          expect.objectContaining({
+            headers: {
+              'Content-Type': 'application/json',
+              Authorization: `Bearer ${token}`,
+            },
+          }),
+        );
 
-      expect(resp).toStrictEqual(expectedResp);
-    });
+        expect(resp).toStrictEqual(expectedResp);
+      },
+    );
 
     it('should fail to get revision data', async () => {
       fetchMock.mockRejectOnce(new Error());


plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
```diff
@@ -114,7 +115,7 @@ describe('ArgoCD service', () => {
       );
 
       const resp = await argoService.getRevisionData(
-        'https://argoInstance1.com',
+        'https://argoinstance1.com',
```
evanshortiss: Constructing URLs with `new URL` forces the resulting URL string to lower case. Technically this line change is unnecessary, but the `expect` on line 152 must be lowercase so making them match helps IMO.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
```diff
@@ -1584,14 +1622,14 @@ describe('ArgoCD service', () => {
...
-        'Error getting token from Argo Instance argoInstance1: FetchError',
+        'Error getting token from Argo Instance argoinstance1: FetchError',
       );
     });
   });
 
   describe('updateArgoProjectAndApp', () => {
     const data: UpdateArgoProjectAndAppProps = {
-      instanceConfig: { name: 'argoInstanceName', url: 'url' },
+      instanceConfig: { name: 'argoInstanceName', url: 'http://argoinstance.com' },
```
evanshortiss: `new URL` expects a valid URL string.

punkle (approved):

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
```diff
@@ -194,7 +196,7 @@ export class ArgoService implements ArgoServiceApi {
     };
 
     const resp = await fetch(
-      `${baseUrl}/api/v1/projects/${projectName}`,
+      new URL(`/api/v1/projects/${projectName}`, baseUrl).toString(),
```
fearphage, line 215: This call to `toString` is not needed. `fetch` [accepts `URL` instances](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#making_a_request). All of these references should be clean up.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
```diff
@@ -101,6 +101,7 @@ describe('ArgoCD service', () => {
     mocked(timer).mockResolvedValue(0);
     jest.clearAllMocks();
     jest.restoreAllMocks();
+    // fetchMock.resetMocks();
```
fearphage: Clean up debug.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
```diff
@@ -127,12 +128,49 @@ describe('ArgoCD service', () => {
       });
     });
 
+    it('should get revision data when the Argo CD instance is configured with a trailing slash', async () => {
+      fetchMock.mockResponseOnce(
+        JSON.stringify({
```
fearphage: If you store the raw value is a variable (like `expected`), you can use this to assert the value later instead of copying and pasting it.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
evanshortiss: I thought so too, but the Jest Mock doesn't seem to like it. I can look into it and see if a new version fixes it.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
evanshortiss: Can do. Was mostly matching existing tests and being a bit lazy 😆

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
evanshortiss: @fearphage confirmed issue with the fetch mock - [source here](https://github.com/jefflau/jest-fetch-mock/blob/master/src/index.js#L123-L142). LMK what you'd like to do.

fearphage: It is supported in `v3.1.0`. Our version of the library is just out of date. https://github.com/jefflau/jest-fetch-mock/pull/193

fearphage (approved):

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
```diff
@@ -127,12 +128,45 @@ describe('ArgoCD service', () => {
...
+        expect.stringMatching(`https://argoinstance1.com/api/v1/applications/${name}/revisions/${revisionId}/metadata`),
+        expect.objectContaining({
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer ${token}`,
+          }
+        }),
+      );
+
+      expect(resp).toStrictEqual(expectedResp);
```
fearphage: Actually this assertion is not needed or at least out of place. You're effectively confirming that `fetchMock` works. 

> Mock this response. Did I get back the response I just mocked? `fetchMock` is working as designed.

They have their own tests for that. I'd just remove this assertion and restore the inline call you had (removing `expectedResp`).

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
evanshortiss: Right. It's not published to npm though.

evanshortiss: Unless someone at Roadie or elsewhere wants to maintain a fork and publish it? The maintainer of the jest mock seems to be off the grid a bit.

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
evanshortiss: Not entirely - it's confirming we got the correct response from `getRevisionData`. The test shouldn't assume that the value returned from `getRevisionData` is identical to the data returned from the mock, since the internals of `getRevisionData` _could_ change at some point. However the prior test can handle this, so maybe this test can simply be:

```js
expect(fetchMock.mock.calls[0][0]).toBe(``https://argoinstance1.com/api/v1/applications/${name}/revisions/${revisionId}/metadata``)
```

And call it a day?

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
fearphage: That's quite unfortunate. I understand. I approved the PR even though my approval doesn't really matter (not a maintainer).

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
fearphage: Ahh, I see. This is a bit noisy to confirm that you're getting the raw result from that response. I usually use unique `Symbol`s for that. 

```typescript
const sentinel: unique symbol = Symbol('there can be only one');
// or more simply
const expected = Date.now();
```

Does this need to be this shape to confirm that you're getting the same response back from the server? It looks like the content is important as the test is currently written. Also the test name does not highlight what's really going on here. Also this is almost a 1-to-1 copy of the previous test, I would make that much more obvious:

```js
it.each([
  'https://argoinstance1.com',
  // comment about trailing slash here perhaps
  'https://argoinstance1.com/',
])('should return revidsion data for host - %s', async (host) => {
  // previous test contents here + augmented with the line in your comment above
});
```

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.service.ts
evanshortiss: Thanks. It's a bummer. Seems like a _very_ popular module, and with recent supply chain attacks it'd be good to know it's in good hands...

plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
fearphage: @evanshortiss Thoughts on using [`it.each`](https://jestjs.io/docs/api#testconcurrenteachtablename-fn-timeout) for the tests to make it apparent that they are identical besides a singular input/character?

evanshortiss: Sure! I'll get around to it today. Was traveling and recovering from jet lag the last few days :)

evanshortiss: All good now?

fearphage: Looks great. Much cleaner and  more clear now. Thanks!

Xantier (approved):

PR #2048, opened 2025-09-16
Fix casing bug for header parsing

sblausten:
When the header for this action is defined as `Content-Type` or `Content-type`, automatic json serialisation of the body does not happen resulting in an unhandled exception. This header should be case insensitive. 

### DIFF
--- a/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
+++ b/plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
@@ -123,12 +123,17 @@ export function createHttpBackstageAction(options: {
 
       let inputBody: Body = undefined;
 
+      const inputHeaders: [string, any][] | undefined =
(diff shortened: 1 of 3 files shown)


plugins/scaffolder-actions/scaffolder-backend-module-http-request/src/actions/run/backstageRequest.ts
```diff
@@ -123,12 +123,17 @@ export function createHttpBackstageAction(options: {
...
         input.body &&
         typeof input.body !== 'string' &&
-        input.headers &&
-        input.headers['content-type'] &&
-        input.headers['content-type'].includes('application/json')
+        inputHeaders &&
+        inputHeaders.find(
+          kv =>
+            kv[0].toLowerCase() === 'content-type' &&
+            kv[1].toLowerCase() === 'application/json',
```
punkle: I think its possible that the header value can be:

application/json
application/something+json
application/json+something

^ I would expect all of these to be json parsable.

punkle (approved):

PR #2062, opened 2025-09-30
Fix undefined issue and error handling in GitHub Pull Requests Statistics card if there is a PR without any commit

christoph-jerolimov:
This PR fixes two issues in the github-pull-requests plugin, esp. the card "GitHub Pull Requests Statistics" when it consumes a PR without any commit.

## Before

image

## Interim version

With the change in `PullRequestsStatsCard.tsx` it shows an ErrorPanel that shows the root cause:

Screenshot From 2025-09-30 22-10-40

## Final version

With the other small changes `getCommitDetailsData` and `calculateStatistics` doesn't crash anymore when there is a PR without any commit.

image

Yep, its an edge case but we actually found that in production. ;)

### DIFF
--- a/.changeset/heavy-experts-train.md
+++ b/.changeset/heavy-experts-train.md
@@ -1,0 +1,6 @@
+---
+'@roadiehq/backstage-plugin-github-pull-requests': patch
+---
+
+- Fix an issue that the GitHub Pull Requests Statistics ignores internal issues and shows undefined instead of an error message to the user.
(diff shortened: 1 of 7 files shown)


christoph-jerolimov:
Will take a look into the linter issue

And yes, this is an edge case if you push later (maybe by accident) the same sha to an open PR.

christoph-jerolimov:
Fixed the linter issue by moving some code

christoph-jerolimov:
Do prettier as well... :smirk:

christoph-jerolimov:
And the latest push should fix the tests. Actually, the tests passed before because the error was ignored.

PR #2072, opened 2025-10-22
resyncAppOnAllArgos fails with "Getting unauthorized" when using tokens

tonyhogsten:
**Problem**

When using the /sync endpoint with appSelector, the sync operation fails with "Getting unauthorized for Argo CD instance" even when valid tokens are configured for all ArgoCD instances.

Root Cause:

In argocd.service.ts, the resyncAppOnAllArgos method calls getArgoToken() with the wrong object type:

```javascript
typescriptconst argoAppResp: findArgoAppResp[] = await this.findArgoApp({
  selector: appSelector,
});
if (argoAppResp) {
  const parallelSyncCalls = argoAppResp.map(
    async (argoInstance: any): Promise<SyncResponse[]> => {
      try {
        const token = await this.getArgoToken(argoInstance); 
```

The argoInstance variable is of type findArgoAppResp:

```javascript
typescript{
  name: string;
  url: string;
  appName: Array<string>;
}
```

But getArgoToken() expects an InstanceConfig:

```javascript
typescript{
  name: string;
  password?: string;
  token?: string;    // ← Missing from findArgoAppResp!
  url: string;
  username?: string;
}
```

Since the findArgoAppResp object doesn't contain the token property, getArgoToken() falls through to username/password authentication, which fails for API-key-only accounts with a 401 Unauthorized error.

Solution
Match the instance name from findArgoApp results with the actual configured instance from this.instanceConfigs before calling getArgoToken():

- [-] Added changeset (run `yarn changeset` in the root) Didn't work? 
- [N/A] Screenshots of before and after attached (for UI changes)
- [N/A] Added or updated documentation (if applicable)

### DIFF
--- a/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
+++ b/plugins/backend/backstage-plugin-argo-cd-backend/src/service/argocd.test.ts
@@ -829,12 +829,6 @@ describe('ArgoCD service', () => {
           ],
         }),
       );
-      // token
-      fetchMock.mockResponseOnce(
-        JSON.stringify({
-          token: 'testToken',
-        }),
-      );
       // sync
       fetchMock.mockResponseOnce(
         JSON.stringify({
@@ -911,12 +905,6 @@ describe('ArgoCD service', () => {
               },
             },
           ],
-        }),
-      );
-      // token
-      fetchMock.mockResponseOnce(
-        JSON.stringify({
-          token: 'testToken',
         }),
       );
       // terminateOperation


tonyhogsten:
This is still relevant.

tonyhogsten:
Its still relevant

punkle:
@tonyhogsten could you look at the failing lint and typescript errors that are being reported by the pipelin

punkle:
@karlhaworth were you able to review this change.

tonyhogsten:
> @tonyhogsten could you look at the failing lint and typescript errors that are being reported by the pipelin

I fixed the error and yarn build and yarn tsc runs cleanly now, my bad.

punkle:
@tonyhogsten could you:

- add a changeset. It can be a revision. To create the changeset run "yarn changeset" from the base directory of the project.
- Update the branch with main

Ill merge when we get those things done. The change looks 👍

tonyhogsten:
@punkle Forgot about that, I've added it. Thank you for your time and looking at it!

punkle (approved):

PR #2079, opened 2025-10-30
update backstage to 1.44.2

punkle:
update backstage to 1.44.2

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -7,8 +7,6 @@
   },
   "bundled": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


plugins/roadie-backstage-entity-validator/package.json
kissmikijr, line 1: I think this cant use backstage:^

punkle: ahh ok. Ill change that, thanks.

Issue #2068, opened 2025-10-16
Support for the new frontend system


drodil:
The new frontend system is ready for adaptation, but all the plugins in this repository are missing support while all plugins in the @backstage/community-plugins already have it. 

## Feature Suggestion

Add support for the new frontend system via `alpha.tsx` exports.

## Possible Implementation

See various examples from the https://github.com/backstage/community-plugins

jar145:
FYI, the Jira plugin still needs to be transitioned over

PR #2092, opened 2025-11-10
Wiz: Add new frontend system support

gaelgoth:
Added support for NFS on Wiz Frontend plugin #2068


.changeset/moody-boats-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@roadiehq/backstage-plugin-wiz': patch
```
Xantier: This should a larger than patch change.

gaelgoth: Done, I changed it to `patch` and fixed format

PR #2104, opened 2025-11-26
Add support for templated tags into provided AWS Lambdas

Xantier:


roadgork (dismissed):

roadgork (approved):

plugins/backend/catalog-backend-module-aws/src/providers/AWSLambdaFunctionProvider.ts
```diff
@@ -139,7 +139,7 @@ export class AWSLambdaFunctionProvider extends AWSEntityProvider {
 
           const entity: Entity | undefined = this.renderEntity(
             {
-              data: lambdaFunction,
+              data: {...lambdaFunction, tags},
```
punkle: The other providers I think use:
{
  data: lambdaFunction,
  tags: tags
}

examples: elasticache, load balancer, account, dynamo db

Xantier: :/ 

That's a different structure than resources that have built-in tags. I'll change this to be equal

PR #2115, opened 2025-12-01
fix(backstage-plugin-argo-cd): display source/destination for multi-source applications

mdcassim:
Display repo URL and path for ArgoCD multi-source applications by falling back to spec.sources[0] when spec.source is absent.

Files changed:
- .changeset/fix-multi-source-details.md
- plugins/frontend/backstage-plugin-argo-cd/src/components/DetailsDrawer.tsx

- Relates to PR #2025 which applied a similar fix to the history card.


mdcassim:
@Xantier, thanks! I have fixed it now.

Issue #1883, opened 2025-03-28
`msw` dependency on production dependencies


Sarabadu:
## Expected Behavior
msw version used for test should not be part of the dependency tree when installing @roadiehq/backstage-plugin-github-pull-requests

## Current Behavior
`msw` is used for testing the packages but at least the plugin @roadiehq/backstage-plugin-github-pull-requests includes it as prod dependency.

this bring the msw version usend on this package when is not necessary and creates maintenance overhead (scan for vulnerabilities, node_modules size)

## Steps to Reproduce

install `@roadiehq/backstage-plugin-github-pull-requests` and run `yarn why msw`

## Possible Solution
move the dependency to devDependencies

## Your Environment

- `@backstage/plugin-catalog-react` version:
- `@backstage/integration` version:

PR #2126, opened 2025-12-15
Move `msw` from prod dependencies to dev dependencies

dotboris:
Fixes: #1883

This PR moves all instances of `msw` from `dependencies` to `devDependencies`. This means that `msw` will no longer show up as a transitive dependency for users of these packages.


punkle:
@dotboris looks good. Just need to fix the linting and we can merge it.

PR #2128, opened 2025-12-15
Support prefixes for argoCD instances on backstage-plugin-argo-cd-backend

srlobo:
Actually the `backstage-plugin-argo-cd-backend` doesn't allow specifying an argoCD instance with a prefix different than `/`. This limitation is not documented anywhere, and since the config item is named `url` I inferred it can be any URL, not only the protocol-host-port part.

This PR fixes that.

#### :heavy_check_mark: Checklist

- [x] Added tests for new functionality and regression tests for bug fixes
- [x] Added changeset (run `yarn changeset` in the root)
- ~~[ ] Screenshots of before and after attached (for UI changes)~~
- ~~[ ] Added or updated documentation (if applicable)~~

PR #2136, opened 2026-01-09
 Support prefixes for argoCD instances on backstage-plugin-argo-cd-backend

srlobo:
(Coming from https://github.com/RoadieHQ/roadie-backstage-plugins/pull/2128).

Actually the `backstage-plugin-argo-cd-backend` doesn't allow specifying an argoCD instance with a prefix different than `/`. This limitation is not documented anywhere, and since the config item is named `url` I inferred it can be any URL, not only the protocol-host-port part.

This PR fixes that.

- ~~[ ] Screenshots of before and after attached (for UI changes)~~
- ~~[ ] Added or updated documentation (if applicable)~~

### DIFF
--- a/plugins/backend/catalog-backend-module-okta/README.md
+++ b/plugins/backend/catalog-backend-module-okta/README.md
@@ -41,6 +41,7 @@ catalog:
 in Okta.
 
 Please note:
+
 - the app needs to be of type "API services"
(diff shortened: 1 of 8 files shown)


Xantier (approved):
Thanks!

PR #2151, opened 2026-01-22
Use hash to shorten id

Irma12:
Use hashed file path so id doesn't exceed limit of 200 chars

### DIFF
--- a/plugins/backend/glean-backend/src/client/GleanIndexClient.ts
+++ b/plugins/backend/glean-backend/src/client/GleanIndexClient.ts
@@ -25,7 +25,7 @@ import {
   LoggerService,
 } from '@backstage/backend-plugin-api';
 import { TechDocsClient } from './TechDocsClient';
-import { DocumentId, GleanDocument } from './types';
+import { GleanDocument } from './types';
 import { CatalogApi } from '@backstage/catalog-client';
 import crypto from 'crypto';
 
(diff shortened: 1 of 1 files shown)


plugins/backend/glean-backend/src/client/GleanIndexClient.ts
```diff
@@ -88,13 +89,19 @@ export class GleanIndexClient {
...
+      .slice(0, 12);
+
     const partialDocument = {
       container: this.techDocsClient.getEntityUri(entity),
       datasource: this.config.getString('glean.datasource'),
       id: `${this.techDocsClient.getEntityUri(
         entity,
-      )}/${filePath}` as DocumentId,
-      // these permissions allow anyone who can sign in to our Glean instance (via Okta SSO) to view the document
+      )}/${filePathHash}` as DocumentId,
```
punkle: This does not fit the DocumentId.. maybe the type needs fixing:

```
TS2322: Type
`${string & {
  __brand: 'S+/S+/S+';
}}/${string}`
```

is not assignable to type
```
`${`${string & {
  __brand: 'S+/S+/S+';
}}` & {
  __brand: '/.+';
}}`
```

plugins/backend/glean-backend/src/client/GleanIndexClient.ts
```diff
@@ -88,13 +89,19 @@ export class GleanIndexClient {
...
       });
 
+    // Shorten the document id by hashing the filePath so 200 characters limit isn't exceeded
+    const filePathHash = crypto
+      .createHash('sha256')
+      .update(filePath)
+      .digest('hex')
+      .slice(0, 12);
+
     const partialDocument = {
```
punkle:
```suggestion
    const partialDocument: Omit<GleanDocument, 'body'> = {
```

plugins/backend/glean-backend/src/client/GleanIndexClient.ts
punkle: Maybe Documentid needs to be this: "type DocumentId = `${EntityUri}/${string}`;"

Irma12: Yes, good catch, I updated PR

punkle: This casting to DocumentId is not needed now that you have added the type to the document.

Irma12: Done

PR #2245, opened 2026-05-23
feat: iframe srcFromAnnotation transforms

robbat2:
The IFrame plugin had the srcFromAnnotation, which took URLs directly from annotations; however there was no way to use an annotation value as part of an input. This functionality permits easy re-use of existing annotations in new ways, such as validating iframe behaviors against a different remote for staging/development rather than production.

- [N/A] Screenshots of before and after attached (for UI changes)

### DIFF
--- a/.changeset/iframe-annotation-transform-config-factory.md
+++ b/.changeset/iframe-annotation-transform-config-factory.md
@@ -1,0 +1,12 @@
+---
+'@roadiehq/backstage-plugin-iframe': minor
+---
+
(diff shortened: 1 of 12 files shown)


.changeset/iframe-annotation-transform-config-factory.md
Xantier, line 1: Let's combine these into a single changeset

Xantier (approved):

.changeset/iframe-annotation-transform-config-factory.md
Xantier: nvm, it's fine.

PR #2284, opened 2026-08-14
feat(@roadiehq/backstage-plugin-argo-cd-node): Add warning logs for unexpected HTML responses from Argo session api

vanessap-aa:
We want to add two small new features:
- We added a debug level log when the Argo Session API returns an unexpected HTML response.
- We are now throwing intentionally when the response is an html, rather than the error thrown when we try to parse a non-json response.

Reasoning: Depending on the deployment environment, traffic to the Argo API may pass through edge security, web application firewall (WAF), or content delivery network (CDN) services such as Akamai, F5, or Cloudflare. When routing failures occur, these services may return an HTML response instead of the expected API response.

Previously, these responses were truncated in logs (see image below), making it difficult to access diagnostic information — such as reference IDs — that can help identify the source of the failure. This change improves observability by logging unexpected HTML responses, providing additional context for troubleshooting and root-cause analysis. For context, the truncation happens when [we try to parse the html as a json](https://github.com/AmericanAirlines/roadie-backstage-plugins/blob/main/plugins/backend/backstage-plugin-argo-cd-node/src/service/argocd.service.ts#L267).

Screenshot 2026-08-13 at 5 38 33 PM

Behavior: No functional behavior has changed. This change only adds additional warning-level logging for unexpected HTML responses to the session endpoint. Does not log when content type includes application/json

New Log - would include helpful troubleshooting information like reference ids:

Screenshot 2026-08-13 at 5 37 23 PM

Additional Change to TESTS:

I noticed the unit tests were nested under the wrong "describe" so I moved it out just for organizational purposes. You can view the actual changes by **HIDING THE WHITE SPACE**


plugins/backend/backstage-plugin-argo-cd-node/src/service/argocd.service.ts
```diff
@@ -258,6 +258,7 @@ export class ArgoService implements ArgoServiceApi {
           password: password || this.password,
         }),
       });
+      await logIfHtmlResponse(resp, this.logger);
```
punkle: I think it makes sense to wrap line 268 in a new parseJson body function and check the content type and fail if the content type is incorrect.

plugins/backend/backstage-plugin-argo-cd-node/src/utils/logIfHtmlResponse.ts
```diff
@@ -0,0 +1,29 @@
...
+import { LoggerService } from '@backstage/backend-plugin-api';
+
+export async function logIfHtmlResponse(resp: Response, logger: LoggerService) {
+  const contentType = resp.headers.get('content-type');
+  if (
+    contentType?.includes('text/html') &&
+    !contentType?.includes('application/json')
+  ) {
+    const body = await resp.clone().text();
+    logger.warn(
```
punkle: This should be debug at most. Otherwise its possible to bloat logs with big html pages.

```suggestion
    logger.debug(
```

plugins/backend/backstage-plugin-argo-cd-node/src/service/argocd.service.ts
vanessap-aa: Hey, thanks for the suggestion. It’s currently failing anyway when the response isn’t JSON and parsing is attempted, so this change should be safe for other users and shouldn’t introduce any new behavior.

Let me know if something like this is what you had in mind.
 
```suggestion
      await parseJsonLogHtml(resp, this.logger); // this would return the json body, but also log the html
```

Would you rather the name of the function like this -> `parseJson`

plugins/backend/backstage-plugin-argo-cd-node/src/utils/logIfHtmlResponse.ts
vanessap-aa: Thanks for the suggestion, @punkle . I see it a little differently and don’t think this should be considered a debug statement.

If an API call that is expected to return an API response instead returns HTML, that indicates an unexpected condition that I think should be visible to the community rather than hidden behind debug logging.

HTML responses from WAFs, CDNs, and similar edge services are typically short and focused. They often contain a basic status message such as 403 Forbidden or Access Denied, a brief explanation, and—importantly—a reference ID or tracking code that can be used to troubleshoot the failure.

Surfacing that response provides useful context for identifying infrastructure or routing issues that would otherwise be difficult to diagnose.

My only concern with relying on debug logging is that enabling debug logs for a Backstage instance in production would surface significantly more logging overall than this individual warning would.

plugins/backend/backstage-plugin-argo-cd-node/src/service/argocd.service.ts
punkle: Yeah I think we should call the function parseJson and add error handling and logging inside if the response does not have the application/json content type. And the replace the "response.json()" function call with your new function. Make it a private function so that you will not need to pass around the logger. If the content does not say the body is json.. there is no need to attempt to parse the body.

plugins/backend/backstage-plugin-argo-cd-node/src/utils/logIfHtmlResponse.ts
punkle: There is a couple of arguments for not logging the body of the response by default:

- We dont know what it is the body of the response. Any proxy server, middleware or http server can intercept the request and return html with anything.. we may be reasonably confident today that the WAFs, CDNS etc that we use do not contain sensitive information.. but that may not hold into the future.
- We dont control where the logs end up. The logging systems have a different authorization platform and trust tier. e.g. administrators debugging the backstage application should not neccessarily have access to the application level data being passed back and forth. If sensitive information is passed in the body of the response, it will appear in the logs and this trust plane is different to how permissions are applied in the application.
- If sensitive data ends up in the logs it is difficult/impossible to remove. It may have reached many downstream systems: kubernetes logs, file systems, s3/cloudtrail, some external logging pipelines, reporting systems.

I believe we have two options to proceed:

- remove the body from the warn log line
- change this log line to debug (so the default is not to log application data)

plugins/backend/backstage-plugin-argo-cd-node/src/service/argocd.service.ts
vanessap-aa: Okay. I will work on this.

plugins/backend/backstage-plugin-argo-cd-node/src/utils/logIfHtmlResponse.ts
vanessap-aa: I will change it to a debug line. Would you be opposed if we, in addition to the debug log, throw the body? So that the caller of that api can receive that html body in the response?

vanessap-aa:
I’ll work on the suggested changes, but I won’t be able to get to them until Friday. Just wanted to set expectations for when this PR will receive an update. I still have one pending question on one of the comments.

vanessap-aa:
I want to recreate this pr using my personal account, rather than my work account, so that I can build my profile. I am going to close this for now, but reference this pr in my new pr under my personal account name. 

This is the new pr - recreated - https://github.com/RoadieHQ/roadie-backstage-plugins/pull/2288

